PR #13, opened 2025-09-15
[feature] Op-level callback

jianyingzhu:
Add op-level callback for local get/put


flexkv/cache/cache_engine.py
```diff
@@ -301,7 +311,7 @@ def get(self,
         # ignore the last incomplete block
         aligned_length = (token_ids.shape[0] // self.tokens_per_block) * self.tokens_per_block
         aligned_token_ids = token_ids[:aligned_length]
-        token_mask = token_mask[:aligned_length]
+        token_mask[aligned_length:] = False
```
linhu-nv: this was an bug and fixed here (https://github.com/taco-project/FlexKV/pull/6/files#diff-2bb648cd82fe83b02e6b5c70bb2734fc22e4a3c1cf5479213a366a683ad5a131L655), shouldn't be reverted. Please be careful when you rebase to dev. Thanks.

PR #25, opened 2025-10-21
[feature]Add GDS (GPU Direct Storage) Support

wenpengw-nv:
This PR adds GPU Direct Storage (GDS) support to FlexKV, enabling direct data transfer between GPU and SSD.

1. GDS Integration
- **Put Path (GPU → CPU → SSD)/Get Path(SSD->GPU)**:To release GPU resources as quickly as possible, put data use a two-step approach: GPU → CPU → SSD. This allows GPU memory to be freed after the CPU copy completes, without waiting for slower disk I/O. Get data operations leverage direct GDS → GPU transfer
- Main modifications to Cache Engine (`flexkv/cache/cache_engine.py`) and addition of **GDS Transfer Workers**
- **Functionally Ready:** This PR provides a complete, working implementation of GDS support. Further performance tuning and optimization needed.

2. Document
- **User Guide** for configuration requirements for GDS (privileged mode, device mounting)
- Provided reference Docker run command with necessary capabilities

### DIFF
--- a/csrc/gds/tp_gds_transfer_thread_group.cpp
+++ b/csrc/gds/tp_gds_transfer_thread_group.cpp
@@ -34,15 +34,45 @@ TPGDSTransferThreadGroup::TPGDSTransferThreadGroup(
     }
   }
 
+  queues_.resize(num_gpus_);
+  mtxs_   = std::vector<std::mutex>(num_gpus_);
+  cvs_    = std::vector<std::condition_variable>(num_gpus_);
+
   // Create CUDA streams for each GPU
   streams_.resize(num_gpus_);
   for (int i = 0; i < num_gpus_; ++i) {
     cudaSetDevice(dp_group_id_ * num_gpus_ + i);
     cudaStreamCreate(&streams_[i]);
   }
+
+  // Create the thread pool
+  stop_pool_ = false;
+  for (int i = 0; i < num_gpus_; ++i) {
+    threads_.emplace_back([this, i]() {
+      int device_id = dp_group_id_ * num_gpus_ + i;
+      cudaSetDevice(device_id);  // only once
+
+      while (true) {
+        Task task;
+        {
+          std::unique_lock<std::mutex> lk(mtxs_[i]);
+          cvs_[i].wait(lk, [&]{ return stop_pool_ || !queues_[i].empty(); });
+          if (stop_pool_ && queues_[i].empty()) return;
+
+          task = std::move(queues_[i].front());
+          queues_[i].pop();
+        }
+        task();  // 
+      }
+    });
+  }
 }
 
 TPGDSTransferThreadGroup::~TPGDSTransferThreadGroup() {
+  stop_pool_ = true;
+  for (auto& cv : cvs_) cv.notify_all();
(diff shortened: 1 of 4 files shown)


csrc/gds/tp_gds_transfer_thread_group.cpp
```diff
@@ -0,0 +1,149 @@
...
+    const int layer_granularity, 
+    const bool is_mla) {
+
+  std::atomic<bool> failed{false};
+  std::string error_msg;
+  threads_.clear();
+  threads_.reserve(num_gpus_);
+
+  for (int i = 0; i < num_gpus_; ++i) {
+    threads_.emplace_back([&, i]() {
```
linhu-nv: the latest "tp_transfer_thread_group.cpp" now doesn't create dynamic threads in each transfer, instead, it can create a thread pool in the initilization, and then reuse all threads created before, which is much faster. Can you please help to refactor this as "tp_transfer_thread_group.cpp" does? thanks.

flexkv/kvtask.py
```diff
@@ -82,8 +82,8 @@ def __init__(self,
             raise ValueError("enable_cpu must be True")
         if cache_config.enable_remote and not cache_config.enable_ssd:
             raise ValueError("enable_ssd must be True if enable_remote is True")
-        if not cache_config.enable_cpu and not cache_config.use_gds:
-            raise ValueError("use_gds must be True if enable_cpu is False")
+        if not cache_config.enable_cpu and not cache_config.enable_gds:
+            raise ValueError("enable_gds must be True if enable_cpu is False")
```
linhu-nv, line 86: maybe we can add an assertion that enable_ssd and enable_gds cannot be used at the same time

setup.py
```diff
@@ -27,19 +27,23 @@ def get_version():
...
     "csrc/tp_transfer_thread_group.h",
+    "csrc/gds/tp_gds_transfer_thread_group.h",
     "csrc/transfer_ssd.h",
     "csrc/radix_tree.h",
+    "csrc/gds/gds_manager.h",
 ]
 
-extra_link_args = ["-lcuda", "-lxxhash", "-lpthread", "-lrt", "-luring"]
-extra_compile_args = ["-std=c++17"]
+extra_link_args = ["-lcuda", "-lxxhash", "-lpthread", "-lrt", "-luring", "-lcufile"]
```
linhu-nv: We should allow flexKV to run without GDS and GDS depencencies such as cufile. So, maybe we can also decide whether to add gds-related compile or link flags according to a environmental var such as "ENABLE_GDS". Only when this env var is true, we add those flags.

csrc/gds/tp_gds_transfer_thread_group.cpp
wenpengw-nv: Already refactored to thread pool pattern as tp_transfer_thread_group.cpp. Please review.

flexkv/kvtask.py
wenpengw-nv: Done.

setup.py
wenpengw-nv: Done. Added FLEXKV_ENABLE_GDS environment var

PR #29, opened 2025-10-28
Support SGLang

jianyingzhu:
Add SGLang connector, adjust transfer kernel to meet with SGLang layout


linhu-nv:
close as we have another Patch-only PR, which is more up-to-date. Thanks.

PR #38, opened 2025-11-06
[Feature] Enhance cache eviction with frequency-aware grace time mechanism

charliecgxu:
## Problem with LRU-only Eviction
Traditional LRU cache eviction suffers from several critical limitations:
1. **Recency bias**: LRU only considers most recent access, ignoring access frequency
2. **Cache pollution**: One-time bulk operations can evict frequently used items
3. **Weak against scans**: Sequential scans displace entire cache despite low reuse
4. **No frequency discrimination**: Cannot distinguish between occasionally used vs hot items

## LFU-inspired Solution Advantages
This implementation introduces a grace time mechanism with hit rewards that combines LRU recency with LFU frequency awareness:

1. **Frequency sensitivity**: Each cache hit extends grace period by hit_reward_seconds, rewarding frequently accessed items
2. **Scan resistance**: Sequential accesses don't get disproportionate time extensions
3. **Temporal locality preservation**: Maintains LRU's advantage for recent accesses
4. **Adaptive prioritization**: Naturally promotes hot items while allowing cold items to expire

## Technical Implementation
- Replaces last_access_time with grace_time in both Python and C++ implementations
- Adds hit_reward_seconds configuration to control frequency weighting
- Grace time extends on hits: current_time + reward for cold items, cumulative for hot items
- Maintains O(1) time complexity for cache operations

### DIFF
--- a/csrc/bindings.cpp
+++ b/csrc/bindings.cpp
@@ -196,7 +196,7 @@ PYBIND11_MODULE(c_ext, m) {
 #endif
 
   py::class_<flexkv::CRadixTreeIndex>(m, "CRadixTreeIndex")
-      .def(py::init<int, int>())
+      .def(py::init<int, int, int>())
       .def("is_empty", &flexkv::CRadixTreeIndex::is_empty)
       .def("reset", &flexkv::CRadixTreeIndex::reset)
       .def("lock", &flexkv::CRadixTreeIndex::lock, py::arg("node"))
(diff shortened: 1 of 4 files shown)


linhu-nv (approved):
This makes sense to me. Good to merge. 
Besides, I have a quick question, how should we determine the `hit_reward_second`? I think if we want to keep the eviction as we expect, the `hit_reward_second` may need to be tuned according to QPS, mem_pool size, or even dataset characters.

PR #44, opened 2025-11-14
Support construct TensorSharedHandle directly from CUDA IPC Handle

Luis-xu:
Now we can support three different type of data to construct the TensorSharedHandle:

1.  If data is a tensor that is managed by torch, we will use the reduce_tensor method to export the TensorSharedHandle.
2.  If data is a tensor that is allocated by cudamalloc, we will use the cudaIpcGetMemHandle method to export the TensorSharedHandle.
3.  If data is bytes-like, it means the memory has already been shared by CUDA IPC, we will skip the export process to construct the TensorSharedHandle.

### DIFF
--- a/tests/test_memory_handle.py
+++ b/tests/test_memory_handle.py
@@ -1,0 +1,458 @@
+"""
+Test TensorSharedHandle with three different input types:
+1. torch.Tensor (default PyTorch IPC)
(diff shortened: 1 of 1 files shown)


peaceforeverCN:
LGTM

peaceforeverCN:
LGTM

PR #78, opened 2025-12-10
[feature] bind multiple get requests into one batch

linhu-nv:
Now we can bind multiple get requests into one batch, through `kvmanager.launch(..., as_batch=True)` to sync/wait and execute more efficiently. Also get prepared for layerwise transfer.


peaceforeverCN (approved):
lgtm

PR #94, opened 2026-01-27
[feature] Support Multi Instance

wenpengw-nv:
Support Multi worker with the same tp size and one flexkv server to share kv cache.


linhu-nv (approved):
LGTM

PR #103, opened 2026-02-06
Fix the bug of vllm_v1_adapter when the vllm version is higher than 0…

feiqiangs:
适配最新版本的vllm ,修复与最新版本vllm 适配过程中测试遇到的错误


linhu-nv:
Looks good to me. thanks!

PR #108, opened 2026-02-12
[fix] use raw pointer instead of tensors as c++ binding parameter

linhu-nv:
We used to direction use gpu/cpu tensors as a parameter from python to c++ calls. This occationally report "RuntimeError: Cannot access data pointer of Tensor that doesn't have storage" error. Now we directly use raw pointer as a parameter to avoid this.


peaceforeverCN (approved):
lgtm

PR #121, opened 2026-03-17
[bugfix] fix match local bug

wenpengw-nv:
When `index_accel=True (default)`,` CacheEngineAccel.match()` returns `MatchResultAccel `without setting` matched_pos`, leaving it as None. In` _get_impl_local()`, the CPU cache insert condition checks `matched_pos == "local"`, which is always False when matched_pos is None. As a result, data loaded from SSD into CPU staging buffers is always freed **instead of being inserted into the CPU cache.**


zhuofan1123 (approved):

peaceforeverCN (approved):
LGTM

PR #142, opened 2026-04-10
add a  new nixl worker 

teeebin:
add a  new nixl worker to do local transfer.
flexkv/transfer/worker.py
flexkv/transfer/nixlutil.py


flexkv/transfer/nixlutil.py
```diff
@@ -0,0 +1,365 @@
...
+        if not dram_ptr_len or not file_paths:
+            return True
+        path_to_fd, xfer_tuples = _open_xfer_files(
+            file_paths, region_offsets, region_lens
+        )
+        if path_to_fd is None:
+            return False
+        try:
+            reg_file = [(0, 0, path_to_fd[p], p) for p in path_to_fd]
+            if not _nixl_register(self.agent, reg_file, "FILE"):
```
linhu-nv: can we do the register in the initlization of the worker? so that we can skip it in the transfer

teeebin: yes , upload a new version. Move register part into prepare_all_ssd_files ,  prepare_dram_cpu , prepare_vram_gpu in nixlutil.py and move register logic out of _transfer_impl() ,  to line 1471 in worker.py

linhu-nv (approved):

PR #153, opened 2026-04-20
feat: add SLRU eviction policy with strategy pattern refactor

zhjc1124:
## Summary

This PR adds **Segmented LRU (SLRU)** as a new eviction policy and refactors the eviction-comparison logic from a `switch`-case in `CRadixNode::Compare` into a dedicated `IEvictionStrategy` hierarchy.

SLRU resists cache scan pollution by splitting cached nodes into two logical segments:
- **Probationary**: `hit_count < protected_threshold` (default `2`). Newly inserted nodes start here.
- **Protected**: `hit_count >= protected_threshold`. Nodes promoted after being accessed enough times.

During eviction, Probationary nodes are evicted first; within the same segment, classic LRU (`last_access_time`) ordering applies. This is particularly useful for mixed workloads where a few hot prefixes (e.g. system prompts) coexist with large volumes of one-shot queries.

## Why the strategy-pattern refactor

The previous `Compare` did a `switch` on `EvictionPolicy` and combined with a `double get_priority()` that could not cleanly express multi-dimensional keys (SLRU's priority is the pair `(is_protected, last_access_time)`, not a flat double). Replacing it with `IEvictionStrategy` removes magic numbers / bit tricks and makes adding future policies trivial.

### C++ core
- New `csrc/eviction_strategy.{h,cpp}`: abstract `IEvictionStrategy` + 6 concrete strategies (`LRU / LFU / FIFO / MRU / FILO / SLRU`) + factory `create_eviction_strategy(policy, protected_threshold)`.
- `CRadixNode::Compare::operator()` now delegates to `strategy->compare(a, b)`.
- `double get_priority()` replaced by `std::string priority_repr(node)` on the strategy (supports multi-dimensional keys).
- `EvictionPolicy` enum moved into `eviction_strategy.h`.
- `update_time()` saturates `hit_count` at `INT_MAX` to prevent int overflow from silently demoting long-lived hot nodes back to Probationary.
- `CRadixTreeIndex` now stores only `strategy_` as the single source of truth; redundant `eviction_policy` / `protected_threshold` fields and their unused getters are removed.

### Python layer
- `RadixTreeIndex._get_eviction_priority` gains the `"slru"` branch with key `(is_protected, last_access_time)`.
- `CacheEngine` / `CacheEngineAccel` accept a new `protected_threshold` kwarg with validation (`int >= 1`); the eviction-policy whitelist is extended with `"slru"`.
- `GlobalCacheEngine` reads `FLEXKV_SLRU_PROTECTED_THRESHOLD` from env and forwards it into every per-device `CacheEngine`. All 6 call sites were migrated to keyword arguments for robustness against future signature drift.

### Bindings / build
- `pybind11` bindings for `CRadixTreeIndex` and `LocalRadixTree` accept `eviction_policy` + `protected_threshold` with sensible defaults (`"lru"`, `2`).
- `setup.py` picks up `csrc/eviction_strategy.{h,cpp}`.

### Docs
- `docs/eviction_policy/README_{en,zh}.md`:
  - SLRU added to the policy table + per-policy parameters table (`slru_protected_threshold` / `FLEXKV_SLRU_PROTECTED_THRESHOLD`).
  - Dedicated SLRU section describing segments and best-fit workloads.
  - Note: `hit_reward_seconds` has no effect on SLRU (SLRU orders same-segment nodes by `last_access_time`, while `hit_reward_seconds` only affects `grace_time`).

### Tests (`tests/test_cache_engine.py`)
- SLRU added to all parameterized policy tests (creation, consecutive eviction, batch eviction, valid-creation).
- New SLRU-specific tests:
  - `test_slru_protected_node_retained`: Protected nodes outlive Probationary ones regardless of access time.
  - `test_slru_batch_eviction_cross_segment`: batch eviction drains all Probationary before touching Protected.
  - `test_slru_custom_protected_threshold`: custom threshold correctly classifies nodes; same-segment LRU wins the tie.
  - `test_slru_threshold_one_promotes_on_first_hit`: `protected_threshold=1` boundary — a single hit promotes to Protected.
  - `test_slru_all_protected_falls_back_to_lru`: when every candidate is Protected, eviction falls back to same-segment LRU.
- `test_config_init` parameterization extended with invalid `protected_threshold` cases (`0`, `-1`, `1.5`, `None`).

## Backward compatibility

- All existing policies (`lru / lfu / fifo / mru / filo`) keep their exact previous semantics.
- `protected_threshold` is a new optional parameter with default `2`; existing callers that don't pass it are unaffected.
- Public C++ type hierarchy change: `CRadixTreeIndex` no longer exposes `get_eviction_policy()` / `get_protected_threshold()`. A grep across the repo confirmed no external caller used them. If downstream users do, `get_strategy()->priority_repr(node)` covers the diagnostic use case.

## Configuration

```bash
export FLEXKV_EVICTION_POLICY=slru
export FLEXKV_SLRU_PROTECTED_THRESHOLD=2   # optional, default 2
```

Or in YAML:

```yml
eviction_policy: slru
slru_protected_threshold: 2
```

## Test plan

- `pytest tests/test_cache_engine.py` — 183 tests all pass, covering both `CacheEngine` (pure Python) and `CacheEngineAccel` (C++ accel) paths.

### DIFF
--- a/csrc/eviction_strategy.cpp
+++ b/csrc/eviction_strategy.cpp
@@ -1,0 +1,161 @@
+#include "eviction_strategy.h"
+#include "radix_tree.h"
+
+#include <sstream>
+#include <stdexcept>
+
+namespace flexkv {
+
+// ===========================================================================
+// parse_eviction_policy
+// ===========================================================================
+
+EvictionPolicy parse_eviction_policy(const std::string &name) {
+  if (name == "lru")
+    return EvictionPolicy::LRU;
+  if (name == "lfu")
+    return EvictionPolicy::LFU;
+  if (name == "fifo")
+    return EvictionPolicy::FIFO;
+  if (name == "mru")
+    return EvictionPolicy::MRU;
+  if (name == "filo")
+    return EvictionPolicy::FILO;
+  if (name == "slru")
+    return EvictionPolicy::SLRU;
+  throw std::invalid_argument(
+      "Unknown eviction policy: '" + name +
+      "'. "
+      "Supported policies: 'lru', 'lfu', 'slru', 'fifo', 'mru', 'filo'.");
+}
+
+// ===========================================================================
+// LRUStrategy
+// ===========================================================================
+
+bool LRUStrategy::compare(CRadixNode *a, CRadixNode *b) const {
+  return a->get_time() > b->get_time();
+}
+
+std::string LRUStrategy::priority_repr(CRadixNode *node) const {
+  return std::to_string(node->get_time());
+}
+
+// ===========================================================================
+// LFUStrategy
+// ===========================================================================
+
+bool LFUStrategy::compare(CRadixNode *a, CRadixNode *b) const {
+  if (a->get_hit_count() != b->get_hit_count()) {
+    return a->get_hit_count() > b->get_hit_count();
+  }
+  return a->get_last_access_time() > b->get_last_access_time();
+}
+
+std::string LFUStrategy::priority_repr(CRadixNode *node) const {
+  std::ostringstream oss;
+  oss << "(" << node->get_hit_count() << ", " << node->get_last_access_time()
+      << ")";
+  return oss.str();
+}
+
+// ===========================================================================
+// FIFOStrategy
+// ===========================================================================
+
+bool FIFOStrategy::compare(CRadixNode *a, CRadixNode *b) const {
+  return a->get_creation_time() > b->get_creation_time();
+}
+
+std::string FIFOStrategy::priority_repr(CRadixNode *node) const {
+  return std::to_string(node->get_creation_time());
+}
+
+// ===========================================================================
+// MRUStrategy
+// ===========================================================================
+
+bool MRUStrategy::compare(CRadixNode *a, CRadixNode *b) const {
+  // MRU: most recently used is evicted first, so the node with the
+  // LARGEST last_access_time has the LOWEST priority (evicted first).
+  // In the max-heap, the node that should be evicted last (smallest
+  // last_access_time) should have higher priority.
+  return a->get_last_access_time() < b->get_last_access_time();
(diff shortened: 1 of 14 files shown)


csrc/eviction_strategy.cpp
```diff
@@ -0,0 +1,160 @@
...
+// ===========================================================================
+
+bool SLRUStrategy::compare(CRadixNode *a, CRadixNode *b) const {
+  bool a_protected = a->get_hit_count() >= protected_threshold_;
+  bool b_protected = b->get_hit_count() >= protected_threshold_;
+
+  // Protected segment nodes have higher priority (evicted later).
+  // Probationary segment (not protected) is evicted first.
+  if (a_protected != b_protected) {
+    return a_protected > b_protected;
```
feiqiangs: bool 和 bool 对比？

flexkv/cache/cache_engine.py
```diff
@@ -65,10 +66,14 @@ def __init__(self,
         if eviction_policy not in _VALID_EVICTION_POLICIES:
             raise ValueError(f"Invalid eviction_policy: '{eviction_policy}'. "
                               f"Supported policies: {sorted(_VALID_EVICTION_POLICIES)}")
+        if not isinstance(protected_threshold, int) or protected_threshold < 1:
```
linhu-nv, line 69: 这里是不是应该驱逐策略是SLRU再检查就可以？

flexkv/cache/cache_engine.py
```diff
@@ -234,10 +240,14 @@ def __init__(self,
...
```
linhu-nv, line 243: 这里也是

linhu-nv:
looks good to me.

linhu-nv (approved):

PR #164, opened 2026-05-12
[bug fix] Fix KV task resource release

wenpengw-nv:
This PR refactors KV task resource release into a two-stage lifecycle.

  Previously, completed tasks could be removed from self.tasks before the caller had a chance to observe the result through `wait() /
  try_wait()`. Because` _update_tasks()` consumes completion events globally, waiting on one group of tasks could complete and release
  another group first, causing later waits to return `NOTFOUND`.

  The new flow separates task completion from task record removal:

  - On task completion, FlexKV marks the task as completed and sheds heavy resources such as `transfer graph, token ids, slot mapping, token mask, and callbacks`.
  - The lightweight task record remains available so `wait() / try_wait()` can still return the final status and return_mask.
  - Once the caller observes the completed task, the task record is fully released from self.tasks.
  - Cancelled tasks are still removed immediately through the normal release path.

  This keeps the original goal of reducing Python GC pressure from completed tasks while preserving the public wait semantics for tasks whose completion was observed indirectly by another wait/update cycle.

### DIFF
--- a/flexkv/kvtask.py
+++ b/flexkv/kvtask.py
@@ -376,7 +376,6 @@ class KVTaskManager:
         task = self.tasks[task_id]
         if not task.is_completed():
             task.status = TaskStatus.CANCELLED
-            task.shed_heavy_resources()
         self._release_task(task_id)
 
     def check_completed(self, task_id: int, completely: bool = False) -> bool:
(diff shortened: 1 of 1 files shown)


linhu-nv (commented):
有一个问题，如果调用方发起了 async task 但从不 wait/try_wait，completed task 的记录会不会永久留在 self.tasks 中导致内存泄漏

wenpengw-nv:
@linhu-nv 应该不会永久停留，会由`ExpiringDict`兜底

linhu-nv (approved):

PR #166, opened 2026-05-13
Validate block id range in Mempool.recycle_blocks

xianweihuihuan:
Mempool.recycle_blocks currently validates ndim and dtype, but out-of-range block ids are not rejected explicitly. In particular, negative ids are accepted by NumPy indexing and can refer to valid entries from the end of the free mask.

### DIFF
--- a/flexkv/cache/mempool.py
+++ b/flexkv/cache/mempool.py
@@ -1,4 +1,5 @@
 from collections import deque
+from operator import le
 from typing import List
 
 import numpy as np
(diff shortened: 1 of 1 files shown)


flexkv/cache/mempool.py
```diff
@@ -1,4 +1,5 @@
 from collections import deque
+from operator import le
```
linhu-nv: 这里似乎没必要引入le, 后面没有用到？

linhu-nv (commented):
感谢贡献！

flexkv/cache/mempool.py
xianweihuihuan: 您说得对，这个 import 是 IDE 自动补全加的，没有用到。已经在最新 commit 删除了，感谢您细致的 review！

linhu-nv (approved):

PR #173, opened 2026-06-01
fix: move critical section updates inside lock in SharedOpPool.allocate_slot

xianweihuihuan:
## Problem

In `SharedOpPool.allocate_slot`, `slot_ref_count` and `slot_hashes` 
updates were outside the lock, creating a race condition where two 
threads could obtain the same slot_id and corrupt the reference count.

## Fix

Move `slot_ref_count[slot_id] += 1` and `slot_hashes[slot_id] = slot_hash`
inside the lock to ensure atomic slot allocation.


linhu-nv (approved):
LGTM. Thanks!

PR #174, opened 2026-06-01
fix: update task.slot_mapping in _set_slot_mapping_impl

xianweihuihuan:
## Problem

In `_set_slot_mapping_impl`, `task.graph` GPU block ids and `task.status`
were updated, but `task.slot_mapping` was not.

This caused `merge_to_batch_kvtask` to read the fake zero slot_mapping
set during `create_get_task`, leading to incorrect batch KV task merging.

## Fix

Add `task.slot_mapping = slot_mapping` after updating the graph,
ensuring the task reflects the real slot mapping before status is
set to READY.


linhu-nv (approved):
that's really a big issue. I am just curious why this is not exposed earlier to us. Anyway thanks for the fix.

xianweihuihuan:
Thanks for the kind words! I noticed it while studying the codebase 
for my own project. Happy to contribute!

PR #175, opened 2026-06-03
Docs inclusive eviction

xuhao77:
## Summary

Clarify FlexKV's multi-level cache eviction behavior.

FlexKV uses an approximately inclusive / write-through cache model across CPU, SSD, and remote storage. This means `put` attempts to populate lower tiers when available, while eviction only removes entries from the current tier and does not trigger write-back or migration.

Also fixes the `CacheEngineAccel.take()` return type annotation to `np.ndarray`, matching the actual mempool allocation result.

## Testing

Not run; documentation and type annotation update only.


linhu-nv (approved):
Looks good to me. Thanks!

PR #184, opened 2026-06-16
refactor: own sub-tasks at batch merge to fix batch task lifecycle leak

xianweihuihuan:
## Problem

When `launch(as_batch=True)` merges sub-tasks into a batch task, the old implementation kept sub-tasks in `self.tasks` and relied on a `pending_sub_count` refcount that decremented only when the caller waited on each sub-task id. But `launch` returns `[batch_id]`, guiding callers to wait on batch_id — which skips the refcount path entirely, leaking all sub-tasks until ExpiringDict expires them after 30 minutes.

## Fix

The engine now takes ownership of sub-tasks at merge time:

- `merge_to_batch_kvtask` pops sub-tasks from `self.tasks` immediately
- Removed `batch_task_id` / `pending_sub_count` and the upward-tracing logic in `_release_task`
- The vLLM adapter now maps `batch_id -> [sub-tasks]` to demux results, and waits on the returned id. The `as_batch=False` path is preserved (each id maps to a single-element list).
- The TRT-LLM adapter is left unchanged: it never enables batching, so it stays on the non-batch path and is unaffected.

## Tests

Added `tests/test_batch_kvtask.py` covering batched GET and PUT: eager-pop, single batch handle, NOTFOUND on merged-away sub-task id, per-sub-task `return_mask` list, and no-leak guarantee.

## Misc

- Updated the dist benchmarks to wait on the id returned by `launch` (instead of the now-popped sub-task ids) and to handle the list-form `return_mask`; also dropped an invalid `layerwise_transfer` kwarg that `launch()` never accepted.
- Added `.venv/` and `ssd_cache*/` to `.gitignore`.


linhu-nv (approved):
LGTM

PR #204, opened 2026-07-07
feat(nvfp4): NVFP4 KV cache support for FlexKV + vLLM integration

xuhao77:
vLLM stores an nvfp4 KV cache as a single packed uint8 tensor whose per-head last dim is head_size//2 + head_size//16 (fp4 data + fp8 block scales inline). FlexKV's transfer/storage core is byte-based, so support needs only a dtype map + a head_size fold so the CPU/SSD mirror matches the packed GPU tensor.

- integration/config.py: map nvfp4/fp4/e2m1 -> uint8 in _parse_dtype_str; add _is_nvfp4_dtype_str + nvfp4_kv_cache_full_dim helpers; fold packed width into head_size in post_init_from_vllm_config (non-MLA guarded, MLA warns).
- integration/config.py: warn + TODO(nvfp4) when nvfp4 is requested via the sglang / trtllm adapters, which do not yet apply the packed head_size fold.
- integration/dynamo/collector.py: TODO(nvfp4) noting Dynamo drives FlexKV via the vLLM adapter (the only verified nvfp4 path).
- common/config.py: document nvfp4 in UserConfig.kv_cache_dtype.
- common/storage.py: fix _kv_dim typo in the LAYERBLOCK branch (blocked all dtypes through the vLLM >=0.23 non-MLA layout, not just nvfp4).
- setup.py: build c_ext for a mainstream multi-arch list (sm80-sm120 + PTX, filtered by local nvcc) so one .so is portable across GPUs.
- tests/test_nvfp4_roundtrip.py: byte-exact nvfp4 offload/reload roundtrip.
- .gitignore: stop tracking uv.lock.

Verified on B200 (sm_100): roundtrip 256/256 bytes match; live vLLM byte-exact integration test reloads nvfp4 KV byte-for-byte across all 28 layers.

The test script integrated with vLLM was not placed in the test folder here. Tests should be self-contained.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -90,3 +90,4 @@ ssd_cache*/
 *cache*/
 benchmarks/nvcomp_benchmarks/inputs
 benchmarks/nvcomp_benchmarks/runs
+uv.lock
--- a/flexkv/common/config.py
+++ b/flexkv/common/config.py
@@ -499,7 +499,7 @@ class UserConfig:
     local_ip: Optional[str] = None
     redis_password: Optional[str] = None
     node_ttl_seconds: Optional[int] = None
-    kv_cache_dtype: Optional[str] = None  # Override kv_cache_dtype when TRT config uses "auto". Supported values: "fp8", "float8", "e4m3", "fp16", "float16", "bf16", "bfloat16", "fp32", "float32"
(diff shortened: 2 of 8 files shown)


linhu-nv (approved):
LGTM

PR #210, opened 2026-07-08
Integrate SWA graph build into Full-KV plans

XingLiu1:
## Summary

Fold the SWA (Sliding Window Attention) control plane into the existing Full-KV GET/PUT plan builders. SWA source/destination selection, slot allocation, eviction, node-mount, and transfer-graph construction now happen inside the same `GlobalCacheEngine` path that owns the Full-KV match — instead of a separate SWA wrapper pass. SWA slots are node-mounted directly on the Full-KV radix tree (no standalone SWA index), so a single forward `match_prefix` resolves both the Full-KV prefix and the deepest ready SWA slot on that prefix.

Target model: DeepSeek-V4-Flash (all-SWA), which runs on `CacheEngineAccel` (the C++ `CRadixTreeIndex` prod path).

### Control-plane integration (`cache_engine.py`)
- Build SWA GET/PUT transfer ops inside the Full-KV local/global plan builders (`_get_impl_*` / `_put_impl_*`) rather than in a separate resolver. GET is gated on `swa_aware=True`; the plain path builds no SWA ops.
- **Removed the node-mount mixin entirely.** Each engine (`CacheEngineAccel`, `CacheEngine`) carries its own SWA helpers (`_alloc_swa_slot`, `_free_swa_slot`, `_drain_unmounted_swa_slots`, `_pin_swa_node`); SWA pool init lives per-tier in `GlobalCacheEngine`. This replaces the earlier `_NodeMountedSWAEngineMixin`, `_resolve_swa_read_source`, and `_probe_swa_source` indirection, all deleted.
- **GET source selection** reads `last_swa_node` / `swa_host_slot` directly off the `MatchResult` from the same forward match (no tree re-walk). Usable end is `min(full, swa)`; the selected source hit must land inside the request window. A snapshot for a deeper trailing window that overshoots the request mask is skipped to the next tier (previously an assert — now a graceful `continue`).

### PUT slot lifecycle — reserve up front, mount on completion
- SWA slots are **allocated fail-fast before graph construction** (the data plane needs the slot id to address bytes). If any required tier (CPU / SSD / REMOTE) cannot allocate, the whole PUT recycles all reserved slots + blocks and returns an empty plan (`_fail_put_before_insert`) rather than degrading to a partial SWA state.
- `set_swa()` is **not** called at insert. Each reserved slot is mounted from a **per-op completion callback** (`_publish_swa_put_slot`) keyed on that slot's own transfer op (`SWA D2H` for CPU, `SWA H2DISK` / `SWA H2REMOTE` for SSD / REMOTE). Full-KV readiness and SWA mount therefore publish independently — neither can expose unfilled SWA bytes. Callbacks compose (`_append_op_callback`) instead of overwriting.
- Allocation locks the matched node while evicting SWA-LRU, so a full-pool eviction can't reclaim the very node this PUT is about to mount onto (`_alloc_swa_slot(protected_node=...)`).

### Completion barrier (`common/transfer.py`)
- `merge_to_batch_graph` no longer picks a single terminal op by priority (which could let a batch report "done" on the Full-KV op alone, before its sibling SWA op finished). It now collects the terminal set — GET `[full H2D, SWA H2D]`, PUT `[full D2H, SWA D2H]` — and inserts a `VIRTUAL` join op depending on all of them via `add_virtual_op_for_multiple_finished_ops`. Mixed GET+PUT terminals are rejected. Write-through ops (SSD/REMOTE) stay outside the barrier (fire-and-forget, published by their own callback); the barrier only guards the GPU source lifetime.

### C++ radix (`csrc/radix_tree.{h,cpp}`, `bindings.cpp`)
- `set_swa()` now asserts `old == -1 || old == slot` instead of silently pushing a different live slot to the freed list; the sanctioned unmount path is `record_freed_swa_slot()`. The merge path unmounts before re-mounting to honor this.
- `match_prefix` publishes `last_swa_node` / `swa_hit_blocks` **only inside `if (current_node->is_ready())`** — the invariant that makes reserve-then-mount safe: an unready node's SWA slot is invisible to matching.
- Root nodes are excluded from SWA-only eviction; `CRadixNode::evictable()` is defined out-of-line (after `CRadixTreeIndex` is complete) so the header stays buildable.
- Exposed `CRadixNode::lock` / `unlock` bindings for the protected-node allocation path.

### Request plumbing
- Propagate `swa_slot_mappings` end-to-end through `server` → `client` → `request` → `kvmanager` → `launch_tasks()`, using explicit keyword arguments to avoid positional drift alongside the Full-KV `slot_mappings`.

### Hierarchy / distributed tier (`hie_cache_engine.py`)
- `HierarchyLRCacheEngine` does not expose the node-mounted SWA match/lifecycle contract, so SWA is now **explicitly disabled** there: `swa_enabled` is `False`, `init_swa` raises `NotImplementedError`, and no host pool is built (previously it built a pool whose slots could never mount — making callers believe SWA was usable). DSv4 is unaffected (it uses `CacheEngineAccel`).

### Tests
- New/expanded control-plane coverage: reserve-then-mount visibility (`test_put_builds_full_plus_swa_store_chain`, `test_put_swa_first_stays_hidden_until_full_kv_is_ready`), SSD write-through mount timing, out-of-window source skip, matched-node kept alive during full-pool eviction, hierarchy engine advertising no SWA, fail-fast on slot exhaustion, RPC launch mapping, and the merge-graph VIRTUAL barrier contract.

## Correctness notes

- **Reserve-then-mount is safe** because visibility is gated twice: a slot is mounted only when its own transfer completes, and even a mounted slot on a not-yet-ready Full-KV node is hidden by the C++ `is_ready` gate in `match_prefix`. So neither "SWA bytes land first" nor "Full-KV readable first" can expose garbage.
- **Fire-and-forget write-through node lifetime**: the SSD/REMOTE mount callback fires from the op's `CompletedOp`, which the scheduler queues before the whole-graph completion sentinel (emitted only on `all_transfer_ops_completed()`), so the mount lands before the node is unlocked.

## Testing

Verified on H20 DeepSeek-V4-Flash-FP8 (TP8, `FLEXKV_ENABLE_SWA_TRANSFER=1`):
- **Control-plane / node-mount unit + smoke suites**: pass (no GPU).
- **GPU byte-exact e2e**: `test_swa_control_plane_e2e`, `test_swa_ssd_staging_e2e`, `test_layerwise_dsv4_multi_group_swa_roundtrip`, `test_swa_dispatch` — all pass, main-KV + SWA roundtrip byte-exact.
- **sglang end-to-end accuracy**: cold → warm(GPU) → `/flush_cache` → restore(FlexKV H2D). Greedy (temp=0) output of the FlexKV-restore path is byte-identical to the warm-GPU reference; server log confirms a real `MAIN H2D (layerwise)` restore after the flush (not a surviving GPU hit).

## Known limitations / follow-ups

- SWA on the hierarchical/distributed (p2p/pcfs) tier is intentionally disabled pending a node-mounted SWA contract in the remote radix index.
- Batch/REMOTE barrier semantics and remaining radix equivalence/compaction details are tracked separately.


flexkv/cache/cache_engine.py
```diff
@@ -88,182 +147,176 @@ def __init__(self,
...
-        self._evict_swa(1)
+        self._evict_swa_slots(1)
         slot = self.swa_pool.allocate()
         return slot if slot is not None else -1
 
-    def _drain_swa_slots(self) -> None:
-        """Return SWA slots freed by tree structural changes to the host pool."""
-        if self.swa_pool is None:
+    def _free_unmounted_swa_slot(self, slot: int) -> None:
+        """Return a slot that is not mounted on any radix node."""
```
Luis-xu: Recycle the slot that is not mounted on any radix node.

flexkv/cache/cache_engine.py
```diff
@@ -88,182 +147,176 @@ def __init__(self,
...
-        for slot in self.index.drain_freed_swa_slots():
-            if slot is not None and slot >= 0:
-                self.swa_pool.free(int(slot))
+        self.swa_pool.free(int(slot))
 
-    def _evict_swa(self, num_swa_evicted: int) -> int:
-        """SWA-only eviction on the node-mounted tree, then drain freed slots and
-        recycle any full blocks freed by leaf deletions. Returns # SWA freed."""
+    def _drain_unmounted_swa_slots(self) -> None:
+        """Return slots detached by radix-tree structural changes to the pool."""
```
Luis-xu, line 190: Get the unmounted slots and recycle

flexkv/cache/cache_engine.py
```diff
@@ -88,182 +147,176 @@ def __init__(self,
...
-        """Like match_swa(lock_for_load=True) but ALSO returns the pinned node.
-
-        The data plane needs the node handle to release the pin (dec_swa_lock_ref)
-        when the SWA H2D completes. Returns ``(swa_hit, slot, node)``;
-        ``(0, -1, None)`` on miss (nothing pinned)."""
+    def _mount_reserved_swa_slot(self, node, slot: int) -> None:
+        if node is None or slot is None or slot < 0:
+            return
+        self._mount_swa_slot(node, slot)
+        self._drain_unmounted_swa_slots()
```
Luis-xu: 这里的drain是什么原因，回收可能产生的unmouted slot?

flexkv/cache/cache_engine.py
```diff
@@ -308,7 +361,9 @@ def insert(self,
                physical_block_ids: torch.Tensor,
                num_insert_blocks: int = -1,
                is_ready: bool = True,
-               match_result: Optional[MatchResultAccel] = None) -> Optional[CRadixNode]:
+               match_result: Optional[MatchResultAccel] = None,
+               swa_store: bool = False,
+               swa_store_visible: bool = True) -> Optional[CRadixNode]:
```
Luis-xu: 这个参数控制什么

flexkv/cache/cache_engine.py
```diff
@@ -1996,198 +2272,66 @@ def match_local_accel(self,
...
+                candidates.append((hit, device_type, mr))
+        if not candidates:
+            return block_mask_start, SWAReadSource()
+
+        candidates.sort(key=lambda item: item[0], reverse=True)
+        for swa_hit, _device_type, _mr in candidates:
+            usable_end = max(block_mask_start, min(block_mask_end, swa_hit))
+            if usable_end <= block_mask_start:
+                continue
+            for _candidate_hit, device_type, mr in candidates:
```
Luis-xu: 这个循环逻辑是不是多余的

flexkv/cache/cache_engine.py
```diff
@@ -1996,198 +2272,66 @@ def match_local_accel(self,
...
-    # late-bind). All are no-ops (empty) unless swa_cache.enabled (i.e.
-    # enable_swa_transfer AND the CPU tier has an SWA host pool).
-    # See deployments/swa_design/08_节点挂载SWA架构.md §10.
+            if hit > 0:
+                candidates.append((hit, device_type, mr))
+        if not candidates:
+            return block_mask_start, SWAReadSource()
+
+        candidates.sort(key=lambda item: item[0], reverse=True)
+        for swa_hit, _device_type, _mr in candidates:
```
Luis-xu: 既然上面进行排序，这里还需要再进行循环？是不是我们拿到排序的最大值就行？

flexkv/cache/cache_engine.py
XingLiu1: Recycle发生在_drain_unmounted_swa_slots里

flexkv/cache/cache_engine.py
XingLiu1: 是没有必要，这个跟tree里的一个setswa的时候，替换一个old swa的逻辑有关系。那块也有点问题，一起修改了

flexkv/cache/cache_engine.py
XingLiu1: 已删除

flexkv/cache/cache_engine.py
XingLiu1: 只有每次recycle后才会drain

flexkv/cache/cache_engine.py
XingLiu1: 已优化

flexkv/cache/cache_engine.py
XingLiu1: 已优化

PR #224, opened 2026-07-15
Moritzxu/swa refactor

Luis-xu:
Resolving reapeted reading and writing causing by swa mount timing.

- Split SWA lifecycle into mount_swa() and publish_swa(). Mounted-but-unpublished slots remain invisible to GETs and outside the SWA LRU.
- Require has_swa && full_ready && swa_ready for SWA cache hits.
- Publish SWA at op completion for PUT and Local GET. Global GET publishes after the full transfer graph completes.
- Handle concurrent GET promotion races using transient CPU staging, ensuring correct GPU data without overwriting the winning mount.
- Skip duplicate SWA writes when PUT loses a mount race.
- Add destination flight locks and source pins to prevent eviction during in-flight transfers.
- Preserve the lock invariant: Full lock before SWA lock, SWA unlock before Full unlock.
- Return individual SWA transfer op IDs so readiness callbacks can attach to the correct DISK2H, REMOTE2H, D2H, or write-through operation.
- Protect pending SWA mounts from Full KV eviction and align Python in_use() behavior with C++.
- Update control-plane, node-mount, staging, lifecycle, and eviction tests.

Known limitation: task cancellation cleanup remains separate work and may still leave locks or transient resources unreleased.

### DIFF
--- a/csrc/bindings.cpp
+++ b/csrc/bindings.cpp
@@ -756,6 +756,8 @@ PYBIND11_MODULE(c_ext, m) {
       .def("get_lock_cnt", &flexkv::CRadixNode::get_lock_cnt)
       .def("lock", &flexkv::CRadixNode::lock)
       .def("unlock", &flexkv::CRadixNode::unlock)
+      .def("is_ready", &flexkv::CRadixNode::is_ready)
+      .def("merge_child", &flexkv::CRadixNode::merge_child)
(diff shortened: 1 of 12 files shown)


Luis-xu:
Discussions indicate that the process of mounting SWA after the transfer completes has minimal impact on overall performance; therefore, this PR will be set to "draft" status.

PR #231, opened 2026-07-22
Merge/dpsk4 mooncake store

Luis-xu:
## Overview

This PR integrates [mooncake-store](https://github.com/kvcache-ai/Mooncake) as a distributed KV cache backend for FlexKV, wired in as the `REMOTE` tier alongside the existing PCFS / P2P remote paths. Unlike the file-addressed 3rd-party remote (block-id → file offset), mooncake-store is **key-addressed**: each KV / SWA block is keyed by its content hash, so the cluster pool is prefix-shareable across all FlexKV instances without a coordinated block-id allocator.

The integration is fully opt-in via `FLEXKV_USE_MOONCAKE_STORE_BACKEND=1` (or `UserConfig.use_mooncake_store_backend=True`) and is orthogonal to the existing GPU / CPU / SSD tiers, layerwise transfer, EAGLE, DP-attention, and DSv4 SWA/state sidecars. When disabled the code paths remain unchanged.

Key design points:
- **Key semantics** — `<block_hash>_FlexKV[_pp_rank_i_of_N]` for main KV, `<tail_hash>_FlexKV_swa` for SWA. PP-rank suffix is elided when the local node covers the full model layer range.
- **PUT/GET reuse** the same `H2REMOTE` / `REMOTE2H` `TransferOp` types; only the payload carries `mooncake_store_block_hashes` (main) or `mooncake_store_swa_block_hashes` (SWA) so the batch-merge and graph-scheduler stay backend-agnostic.
- **Zero-copy** — the CPU-tier pinned buffer is directly registered with mooncake for RDMA, so no extra staging buffer is introduced.
- **Match RPC** — a single `batch_is_exist(kv_keys + swa_keys)` per sequence returns both the KV longest-prefix and the SWA joint hit, avoiding two round-trips.

### 1. New adapter module — `flexkv/external/`

| File | Purpose |
|---|---|
| `mooncake_store_utils.py` | `MooncakeStoreConfig` (JSON loader), `MooncakeStoreClient` (thin `MooncakeDistributedStore` wrapper with `batch_put` / `batch_get` / `batch_exists` / `register_buffer`), and `MooncakeStoreCacheEngine` (drop-in replacement for `CacheEngineAccel` presenting `match()` / `insert()`) |
| `mooncake_store_keys.py` | Single source of truth for key construction (`PoolKind.KV` / `PoolKind.SWA` + `build_key`) shared by client, cache engine, and transfer worker |
| `__init__.py` | Public re-exports with lazy import so unrelated FlexKV runs pay no mooncake import cost |

### 2. `MooncakeStoreTransferWorker` — new remote I/O worker
`flexkv/transfer/worker.py`
- Runs in the same spawn'd subprocess model as other workers; registers the CPU pinned buffer with mooncake and answers `H2REMOTE` / `REMOTE2H` ops
- Two variants: main-KV pool (`PoolKind.KV`) and SWA pool (`PoolKind.SWA`, pure-client mode — `override_global_segment_size=0` — so SWA workers do not contribute their own storage segment)
- `_preprocess_kv` / `_preprocess_swa` compute `(cpu_ptrs, sizes, keys)` from the per-op block-id → CPU-slot mapping and the sequence hashes carried on the op

### 3. Cache-engine wiring — `flexkv/cache/cache_engine.py`
- `remote_cache_engine` is instantiated as `MooncakeStoreCacheEngine` when the flag is on (else keeps the existing `HierarchyLRCacheEngine` / PCFS paths)
- `get_match`/`put_match` now propagate the sequence hashes onto `op_remote2h.mooncake_store_block_hashes` / `op_h2remote.mooncake_store_block_hashes`, using `remote_matched_result.kv_matched_blocks` (added to `MatchResultAccel`) so PUT skips keys that already exist in the store
- SWA path (`_select_swa_read_source` → `_reserve_swa_read`, PUT-side SWA chain in `put_match`) gains a mooncake branch: the tail hash of the SWA-covered prefix becomes the SWA key, and `SWAReadSource` carries `mooncake_tail_hash` in place of a slot / node handle (`is_mooncake` short-circuits the pin / unlock logic which does not apply on a key-addressed tier)

### 4. `TransferOp` / `WorkerTransferOp` — hash fields
`flexkv/common/transfer.py`, `flexkv/transfer/worker_op.py`
- Two optional fields on `TransferOp`: `mooncake_store_block_hashes` (`np.ndarray`, main KV) and `mooncake_store_swa_block_hashes` (`List[str]`, SWA tail hashes)
- `WorkerTransferOp` forwards both to the worker process; when the fields are set, the op-buffer fast path is bypassed so block ids stay accessible

### 5. Batch merge — main and SWA lanes split
`flexkv/common/transfer.py::merge_to_batch_graph`
- Refactored into `_merge_ops` (main-KV) + `_merge_swa_ops` (SWA), each concatenating their respective hash lists in-order and rejecting cross-mixing
- `supported_types` extended with `H2REMOTE` / `REMOTE2H`; batch sink logic factored into `_add_batch_sink` (0/1/N terminals) and `_pick_dp_client_id`
- Fused-layerwise GET path now keeps REMOTE2H (main + SWA) as standalone predecessors of the fused LAYERWISE op instead of folding them in — remote lanes remain SDK-driven while the local lanes stay layerwise-fused
- Enforces "no mixed GET/PUT in a single batch" and "layerwise batches are GET-only" via asserts

### 6. Storage engine — no separate remote region when mooncake is on
`flexkv/storage/storage_engine.py`
- When `use_mooncake_store_backend` is true, the hot CPU / SWA CPU buffers are registered with mooncake directly; the previous "allocate a separate REMOTE contiguous layout" path is skipped for both main and SWA pools

### 7. Transfer engine — worker map wiring
`flexkv/transfer/transfer_engine.py`
- Creates a `MooncakeStoreTransferWorker` for main-KV `H2REMOTE` / `REMOTE2H` (mutually exclusive with the PCFS remote path)
- Creates a SWA-pool variant when SWA is enabled (`_swa_worker_map[H2REMOTE/REMOTE2H]`)

### 8. Config — new user-facing options
`flexkv/common/config.py`
- `UserConfig` / `CacheConfig`: `use_mooncake_store_backend`, `mooncake_store_config_path`, plus five layer-range / PP fields (`mooncake_store_pp_rank`, `mooncake_store_pp_size`, `mooncake_store_node_layer_start`, `mooncake_store_node_layer_end`, `mooncake_store_total_layers`) auto-populated from `rank_info`
- Env fallback: `FLEXKV_USE_MOONCAKE_STORE_BACKEND` / `FLEXKV_MOONCAKE_STORE_CONFIG_PATH`
- `enable_remote = enable_3rd_remote OR use_mooncake_store_backend`; the "remote requires SSD" precondition is relaxed for mooncake (mooncake bypasses SSD)

## Validation
I have tested the accuracy of DPSK-v4-Flash under TP8 and DP8 configurations. The specific methodology involved running the test dataset once on instance 1 and again on instance 2, then observing the evaluation accuracy for both runs. The results are as follows (each dataset was tested twice):
### tp8:
| 数据集 | instance1 | instance2 |
|--------|-----------|-----------|
| AIME | 0.6000 | 0.6333 |
| HumanEval | 0.915 | 0.915 |
### dp8:
| 数据集 | instance1 | instance2 |
|--------|-----------|-----------|
| AIME | 0.6667 | 0.6667 |
| HumanEval |0.933 | 0.933 |


flexkv/external/mooncake_store_utils.py
```diff
@@ -0,0 +1,710 @@
...
+    def unlock(self, node: Any) -> None:
+        pass
+
+    def set_ready(self, node: Any, ready: bool, ready_length: int) -> None:
+        pass
+
+    def insert_and_publish(self, node: Any) -> bool:
+        return True
+
+    # TAKE is skipped as the real index is managed by simm itself
```
linhu-nv: “mooncake” instead of "simm"

flexkv/cache/cache_engine.py
```diff
@@ -852,7 +870,16 @@ def get(self,
             )
 
         return_mask = np.zeros_like(token_mask, dtype=np.bool_)
-        return_mask[block_start_idx* self.tokens_per_block:
+        if temp_cache_strategy.ignore_gpu and temp_cache_strategy.ignore_gds:
+            prefetch_blocks = 0
+            for op in transfer_graph._op_map.values():
+                if op.transfer_type == TransferType.REMOTE2H:
+                    prefetch_blocks += len(op.src_block_ids)
```
linhu-nv, line 880: 这里应该只统计非swa的blocks吧？否则加重复了

flexkv/common/transfer.py
```diff
@@ -782,88 +874,105 @@ def _concat_swa_block_ids(ops: List[TransferOp]) -> Tuple[np.ndarray, np.ndarray
...
+            if merged_h2d_op is not None:
+                get_sinks.append(merged_h2d_op.op_id)
+            if merged_swa_h2d_op is not None:
+                get_sinks.append(merged_swa_h2d_op.op_id)
+            if not get_sinks:
+                for op in (merged_remote2h_op, merged_swa_remote2h_op,
+                           merged_disk2h_op, merged_swa_disk2h_op):
+                    if op is not None:
+                        get_sinks.append(op.op_id)
+                        break
```
linhu-nv: 这里只找到一个op就break，会不会导致没有GPU的时候（也就是prefetch）full kv或者swa相关的kv有的实际上传输还没完成

flexkv/transfer/worker.py
```diff
@@ -3378,3 +3404,153 @@ def get_node_meta(self, node_id: int) -> Optional[NodeMetaInfo]:
...
+        ``dtype.itemsize``. Single-group layouts still use element count ×
+        itemsize.
+        """
+        if cpu_kv_layout.layer_groups is not None:
+            return int(cpu_kv_layout.get_block_stride())
+        return int(cpu_kv_layout.get_elements_per_block() * dtype.itemsize)
+
+    def _transfer_impl(self, cpu_ptrs, block_sizes, keys, transfer_type: TransferType) -> None:
+        if transfer_type == TransferType.H2REMOTE:
+            self.mooncake_client.batch_put(keys, cpu_ptrs, block_sizes)
```
linhu-nv: 应该检查是否正确完成

Luis-xu: Solved in commit 3a6a020954b78ec1cc2a3c15d692fe80d370a9f2. Now when mooncake-store transfer partially completed, will raise flexkv error log here.

flexkv/external/mooncake_store_utils.py
Luis-xu: Solved in commit 3a6a020954b78ec1cc2a3c15d692fe80d370a9f2.

flexkv/common/transfer.py
Luis-xu: Solved in commit  3a6a020954b78ec1cc2a3c15d692fe80d370a9f2.

flexkv/cache/cache_engine.py
Luis-xu: Solved in commit 3a6a020954b78ec1cc2a3c15d692fe80d370a9f2.

linhu-nv (approved):

PR #235, opened 2026-07-26
perf(transfer): dispatch from a dirty set instead of sweeping every graph

Claisenn:
## Problem

`TransferScheduler.schedule()` visited every in-flight graph on every call: `take_ready_ops()` on all of them, then `all_transfer_ops_completed()` on all of them.

Dependencies never cross graphs — `add_dependency()` asserts both ops belong to the same graph — so a graph can only expose new ready ops after one of its **own** ops completes. Sweeping the rest is wasted work, and it makes the cost of one wake-up scale with total in-flight concurrency rather than with how much actually changed.

## Change

Track the graphs dirtied since the last call and drain only those. `flexkv/transfer/scheduler.py` is the only production file touched.

## Measurements

`python benchmarks/microbenchmark_transfer_scheduler.py --baseline` runs the old full-sweep scheduler and the new one in the same process, so the table is reproducible from one command (best of 5 per point, single machine):

| in-flight graphs | full sweep | dirty set |
|---:|---:|---:|
| 1 | 1.61 us | 1.59 us |
| 8 | 5.16 us | 1.61 us |
| 32 | 16.07 us | 1.66 us |
| 128 | 63.66 us | 1.69 us |
| 512 | 266.32 us | 1.84 us |
| 1024 | 532.37 us | 1.98 us |
| 2048 | 1074.83 us | 1.93 us |

Flat in concurrency instead of linear.

### Scope of that claim

I would rather state the limits than have a reviewer find them:

- This is the extreme case, exactly **one** dirty graph per tick. With `k` dirty the cost is `O(k)`, which is the point, but the table is the most favorable shape.
- This is **control-plane Python overhead only**. Whether it shows up end to end depends on the scheduler thread being a bottleneck. It will not be at low concurrency, nor when batch merging keeps the in-flight graph count small, and the same tick also pays `multiprocessing` pickle costs that dominate at modest concurrency.
- I have **not** measured this end to end on a real GPU deployment.

## The failure mode this had to be careful about

The full sweep was self-healing: any state drift got picked up on the next tick. An incremental scheduler is not, so losing a dirty bit means the graph is never revisited and its request hangs.

The drain therefore drops a graph's dirty bit only **after** that graph has been processed, so a raise leaves it queued for the next call. That recovers the dirty bit, **not** the work: ops already collected into `next_ops` are discarded with the exception, and `take_ready_ops()` will not re-emit an op it already flipped to `RUNNING`. The full-sweep scheduler loses those identically — `test_exception_mid_drain_still_loses_already_collected_ops` pins that limit so the guarantee is not overstated.

## One deliberate behavior change

Completing a virtual op now revisits its graph within the same call. The old code deferred the successors to the next external event, which `TransferEngine._scheduler_loop` only produces when an op finishes or a graph arrives — so a graph whose remaining work sat behind a **non-terminal** virtual op could stall.

Every virtual op FlexKV builds today is a terminal sink (`add_virtual_op_for_multiple_finished_ops`, `_add_batch_sink`), so this is latent rather than live. `test_non_terminal_virtual_op_dispatches_successor_same_tick` covers it.

## Testing

`tests/test_transfer_scheduler_incremental.py` (marked `unit`, no GPU, runs in <0.1s) keeps a reference implementation of the old full-sweep scheduler and drives it in lockstep with the new one over random DAGs — with some graphs arriving mid-run, as the engine does it. Both must dispatch the same ops and complete the same graphs on every tick, and every graph must drain.

I mutation-tested the suite; each of the following is killed by at least one test:

- substituting the old scheduler verbatim → only the two intended divergences fail
- dropping the dirty bit before processing instead of after
- removing the virtual-op revisit
- not dirtying on `add_transfer_graph` (67/67 fail)
- not dirtying on a finished op
- removing the `graph is not None` guard

## Notes for the reviewer

- `completed_graph_ids` and `next_ops` now come back in dirty order rather than scheduler insertion order. Ops in different graphs are independent and the only consumer fans both onto queues per element (`TransferEngine._scheduler_loop`), so I believe the order carries no meaning — worth a second opinion.
- The reference full-sweep scheduler is duplicated in the test and in the benchmark (`--baseline`). I kept both and cross-referenced them in their docstrings; happy to drop `--baseline` instead if you would rather not carry the copy.
- `TransferOpGraph.trigger_op()` completes an op outside `schedule()` and so would not dirty its graph. It has no callers today; I noted the standing requirement next to `_dirty_graph_ids` rather than change it here.

### DIFF
--- a/benchmarks/microbenchmark_transfer_scheduler.py
+++ b/benchmarks/microbenchmark_transfer_scheduler.py
@@ -1,0 +1,163 @@
+"""Microbenchmark: TransferScheduler.schedule() cost vs in-flight concurrency.
+
+The transfer scheduler runs on a single thread and is woken once per batch of
+finished ops. What we want to know is whether the cost of one wake-up depends on
+how much actually changed, or on how many requests happen to be in flight.
+
+This drives a fixed amount of real work per tick -- exactly one op of one graph
+completes -- while varying the number of other, idle in-flight graphs. A
+scheduler whose cost tracks real work stays flat; one that sweeps every graph
+grows linearly with concurrency. Note this is the extreme case, exactly one
+dirty graph per tick; with k dirty the incremental cost is O(k), which is the
+whole point.
+
+CPU only: no GPU, no c_ext transfers, no storage. Run with:
+    python benchmarks/microbenchmark_transfer_scheduler.py --baseline
+"""
+from argparse import ArgumentParser
+import time
+
+import numpy as np
+
+from flexkv.common.transfer import (
+    TransferOp,
+    TransferOpGraph,
+    TransferOpStatus,
+    TransferType,
+)
+from flexkv.transfer.scheduler import TransferScheduler
+
+
+class FullSweepScheduler:
+    """The pre-change scheduler, kept here so `--baseline` can print a
+    before/after table from one run instead of asking you to check out the
+    previous revision.
+
+    `tests/test_transfer_scheduler_incremental.py` keeps its own copy, where it
+    serves as the equivalence oracle; keep the two in step.
+    """
+
+    def __init__(self) -> None:
+        self._transfer_graphs = {}
+
+    def add_transfer_graph(self, graph) -> None:
+        self._transfer_graphs[graph.graph_id] = graph
+
+    def schedule(self, finished_ops):
+        for op in finished_ops:
+            if op.graph_id in self._transfer_graphs:
+                self._transfer_graphs[op.graph_id].mark_completed(op.op_id)
+
+        next_ops = []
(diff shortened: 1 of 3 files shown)


linhu-nv:
Thanks a lot for the contribution! very impressive observations. Actually we also saw this bottleneck on some Generation Recommendation scenarios (very high QPS, about 2k-3k). Just curious that do you find this problem also on some extremely high qps? or how do you identify this can be a bottle neck? becasue normally LLM serving will not have that many inflight graphs. Thanks again.

linhu-nv (approved):

zhuofan1123 (approved):

PR #236, opened 2026-07-26
fix(kvtask): roll back plan resources when cancelling a never-launched task

Claisenn:
## Problem

`create_get_task` / `create_put_task` run the cache-engine planner **at creation time**, which already acquires real resources:

- matched radix nodes are **locked** (`get()` / `put()` call `lock_node` on every entry of the plan's `node_to_unlock`),
- **CPU staging blocks** are allocated for SSD-resident fragments (`_get_impl_local`, `take()`),
- new index nodes are inserted with **`is_ready=False`**, and only a completion callback can publish them.

`KVTaskManager._cancel_task`, however, only popped the task from its dicts — it never invoked any release path.

The vLLM adapter cancels never-launched tasks **routinely**: every request that matches but is not scheduled in that step (the auto-cancel path in `_build_connector_meta`), and every preemption (`handle_preemptions` — whose comment says tasks are moved to the cancel list *"to ensure underlying resources are freed"*, an assumption the code did not satisfy).

Each such cancel therefore leaked its staging blocks out of the CPU mempool and left permanently-unevictable locked/unready nodes in the tree. Unready nodes are additionally skipped by later puts of the same prefix (put skips on matched blocks, not ready-matched ones), so those prefixes became lasting hit-rate holes.

**Reproduction** (control-plane only, no GPU): on a 512-block CPU pool with SSD-resident data, 32 cancelled SSD-hit GETs converted the entire CPU pool into unreachable blocks — every subsequent PUT failed permanently (`CPU=0/16`), with 21 locked and 20 forever-unready nodes left in the CPU tree. With this fix the same scenario ends with the pool fully restored, zero locked/unready residue, and all subsequent PUTs succeeding.

## Fix

Thread an abort path from the planner to the cancel site:

- **`get()` / `put()` now return a `TransferPlanHandle`** instead of a bare `partial`. *Calling* it keeps the exact completion behavior — kvtask, benchmarks and the batch-merge list path are unchanged, since the handle is callable. `abort()` rolls the plan back instead. The handle runs at most one of the two, exactly once, so a cancel racing a completion cannot double-release.
- **`_abort_transfer_plan`** unlocks every node the plan locked, removes the nodes this plan inserted unready and recycles their blocks, recycles staging buffers, releases a held SWA read reservation (get) and returns unpublished SWA slots to the host pool (put). Distinguishing inserted-unready nodes from pre-existing matched nodes needs no extra bookkeeping: matched nodes are ready by definition, so the rollback is a guarded no-op for them.
- **`RadixTreeIndex.remove_unready_leaf` / `CRadixTreeIndex::remove_unready_leaf`** implement the rollback at the index level, mirroring `evict()`'s leaf deletion (including the SWA tombstone cascade). The node is removed only while it is still an unready, unlocked, SWA-unlocked **leaf**; if another request split it or inserted below it in the create-to-cancel window, it is left in place — the pre-existing behavior — rather than risk detaching blocks a descendant depends on.
- **`_cancel_task`** aborts tasks whose graph never launched (`UNREADY` / `READY`). `RUNNING` tasks keep the old behavior: their graph is in flight and the completion callbacks will still fire. (Cancelling running transfers is the larger follow-up the adapter TODO already notes.)

## Testing

`tests/test_cancel_task_rollback.py` (marked `unit`, no GPU):

- index-level rollback for **both** engine implementations — the `CacheEngineAccel` variants exercise the new C++ `remove_unready_leaf` wherever the extension is built (they skip locally without it),
- plan-level abort for get and put on both index modes: mempool restored exactly, zero unready residue, and the same prefix is re-insertable afterwards,
- handle complete/abort mutual exclusion (late cancel racing a completion is a no-op),
- `_cancel_task` wiring, and that `RUNNING` tasks are *not* aborted.

Mutation-checked: removing the `_cancel_task` abort call fails the wiring test; making the index rollback a no-op fails four tests.

## Notes for the reviewer

- The lint hunks (line wraps, an unused-name import, one nested-if collapse) are pre-existing `ruff` findings in the touched files; the changed-file lint job fails the PR without them. No behavior change in any of them.
- Residual known gap, unchanged from before: if another request locks or extends the unready node between create and cancel, the rollback backs off and that node stays unready (exactly the pre-fix behavior for a strictly rarer case).
- I could not compile the CUDA extension locally; the C++ `remove_unready_leaf` mirrors the Python implementation line for line and is exercised by the Accel test variants in CI.

### DIFF
--- a/csrc/bindings.cpp
+++ b/csrc/bindings.cpp
@@ -933,6 +933,9 @@ PYBIND11_MODULE(c_ext, m) {
                &flexkv::CRadixTreeIndex::evict),
            py::arg("evicted_blocks"), py::arg("evicted_block_hashes"),
            py::arg("num_evicted"), py::call_guard<py::gil_scoped_release>())
+      .def("remove_unready_leaf",
+           &flexkv::CRadixTreeIndex::remove_unready_leaf, py::arg("node"),
+           py::arg("freed_blocks"), py::call_guard<py::gil_scoped_release>())
       .def("total_cached_blocks", &flexkv::CRadixTreeIndex::total_cached_blocks)
       .def("total_unready_blocks",
            &flexkv::CRadixTreeIndex::total_unready_blocks)
--- a/csrc/radix_tree.cpp
+++ b/csrc/radix_tree.cpp
@@ -736,4 +736,37 @@ CRadixTreeIndex::match_prefix(torch::Tensor &block_hashes, int num_blocks,
       last_swa_node, swa_hit_blocks);
 }
 
+// Roll back a not-yet-ready inserted leaf (planned transfer aborted before
+// launch). Only removes the node when unambiguously safe: it must still be an
+// unready, unlocked, SWA-unlocked leaf. If another request split it or
+// inserted below it in the create-to-cancel window, it is left in place
+// (pre-existing behavior) rather than risk detaching blocks another plan
+// references. Mirrors flexkv/cache/radixtree.py::remove_unready_leaf.
+int CRadixTreeIndex::remove_unready_leaf(CRadixNode *node,
+                                         torch::Tensor &freed_blocks) {
+  if (node == nullptr || is_root(node) || node->is_ready() ||
+      node->get_lock_cnt() > 0 || node->get_swa_lock_ref() > 0 ||
+      !node->is_leaf()) {
+    freed_blocks.resize_({0});
+    return 0;
+  }
+  std::vector<int64_t> freed;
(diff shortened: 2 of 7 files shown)


linhu-nv (approved):

linhu-nv:
@Luis-xu please help check the cancle logic of swa node, if you have time. Thanks.

Luis-xu (commented):
LGTM! In my view, this pr can not cover the split scenario either (the roll-back of radixNode will be very complex).

PR #237, opened 2026-07-26
fix(transfer): propagate transfer failures instead of silently dropping them

Claisenn:
> **Stacked on #236** — the first two commits here are #236 (never-launched-task rollback); this PR's own changes start at `d2f5e5b`. Review is easiest after #236 merges, or commit-by-commit.

## Problem

A worker that cannot execute a transfer op logs the error and reports nothing — `worker.py`'s own comment: *"only put the op when transfer success"*. The consequences chain all the way up:

1. The op stays `RUNNING` forever, so its graph never completes and sits in the scheduler permanently.
2. The task's `wait()` returns a misleading **TIMEOUT** after 20s — with **zero cleanup**. The caller cannot distinguish slow from broken.
3. Every resource the plan acquired leaks for good: locked radix nodes, `is_ready=False` index nodes (permanently unevictable, and they shadow later puts of the same prefix — lasting hit-rate holes), and CPU staging blocks gone from the mempool.

One SSD read error or remote-store hiccup is enough to poison the tree. `TaskStatus.FAILED` / `KVResponseStatus.FAILED` and their response mapping already existed — no path ever set them. This PR wires the path end to end.

## Design

- **Worker**: reports a failed op as `(op_id, False)` on `finished_ops_queue`. A bare int still means success, so the format stays compatible.
- **Engine loop**: failures route through the same accounting as completions (replica maps, pending counts, pin buffer, nvtx) so nothing leaks, but the op never reaches `finished_ops` — its successors must not run. `scheduler.fail_graph` drops the graph so no remaining op dispatches. Ops with a failed PP replica are discarded, not finalized, when their pending count drains.
- **Drain-then-abort**: the graph-level failure (`CompletedOp.failed_graph`, a new default-`False` flag — pickle- and constructor-compatible) is emitted only once every already-dispatched op of the graph has drained. Late sibling completions therefore still run their op callbacks against index nodes that are still mounted; the rollback never races them.
- **Task layer**: terminates the task as `FAILED` and runs the plan's abort (the `TransferPlanHandle` from #236). Nodes whose transfers completed stay published — their data is valid; nodes whose transfers never ran are removed and their blocks recycled. The same `is_ready` discriminator #236 introduced handles the partial-completion split with no extra bookkeeping.
- **Multi-handle deployments** (`nnodes > 1`, trtllm-remote): the same graph goes to every handle, so graph-level messages aggregate — the task terminates only when **all** handles report a terminal state, any failure wins, and a failed graph's stale per-op dedup counters are purged (they could never reach N). Single-handle deployments (the default) behave exactly as before: one terminal message, immediate effect. Adversarial review caught this one: passing the first failure straight through would have recycled plan blocks a sibling engine was still writing into.

## Known limits, stated rather than hidden

- A worker that **crashes without reporting** still leaves the graph outstanding; the 20s `wait()` timeout continues to cover that case, unchanged.
- A LAYERWISE mid-transfer failure fails the graph and rolls the task back, but an inference thread blocked on a per-layer eventfd still relies on its own timeout (pre-existing behavior, unchanged).
- The lint hunks (line wraps, `l` → `layout`, one nested-if collapse, in-dict-keys) are pre-existing ruff findings in touched files; the changed-file lint job fails the PR without them. `worker.py`'s long-standing below-code import group gets a per-file `E402` acknowledgment in `pyproject.toml` rather than a risky reorder.

## Verification

`tests/test_transfer_failure_propagation.py` (marked `unit`, no GPU, <2s): message shape + pickle/legacy compatibility, `fail_graph` semantics (stop dispatch, never complete, other graphs unaffected, idempotent), the engine failure path including PP-replica drain and discard-not-finalize routing, drain-gated emission, `FAILED` terminal state with full rollback, partial completion keeping the completed tier, `_update_tasks` routing, failure/completion mutual exclusion, and the multi-handle aggregation rules (wait-for-every-handle, failure-wins, stale-counter purge with an unrelated-graph survivor, unchanged single-handle fast path).

Mutation-checked: neutralizing `fail_graph`, the abort call, the drain gate, or the discard-vs-finalize routing each fails its targeted test. Two independent adversarial review rounds ran probes against the loop ordering (VIRTUAL dispatch, FIFO end to end, failure racing completion, batch tasks) and found one real defect (the multi-handle bypass, fixed in `1e20aa5`) and nothing else; existing unit/smoke suites are byte-identical to `upstream/main`.

The accel-index test variants exercise the C++ path where the extension is built (they skip locally); I cannot compile the CUDA extension on my machine.

### DIFF
--- a/benchmarks/microbenchmark_transfer_scheduler.py
+++ b/benchmarks/microbenchmark_transfer_scheduler.py
@@ -1,0 +1,163 @@
+"""Microbenchmark: TransferScheduler.schedule() cost vs in-flight concurrency.
+
+The transfer scheduler runs on a single thread and is woken once per batch of
+finished ops. What we want to know is whether the cost of one wake-up depends on
+how much actually changed, or on how many requests happen to be in flight.
+
+This drives a fixed amount of real work per tick -- exactly one op of one graph
+completes -- while varying the number of other, idle in-flight graphs. A
+scheduler whose cost tracks real work stays flat; one that sweeps every graph
+grows linearly with concurrency. Note this is the extreme case, exactly one
+dirty graph per tick; with k dirty the incremental cost is O(k), which is the
+whole point.
+
+CPU only: no GPU, no c_ext transfers, no storage. Run with:
+    python benchmarks/microbenchmark_transfer_scheduler.py --baseline
+"""
+from argparse import ArgumentParser
+import time
+
+import numpy as np
+
+from flexkv.common.transfer import (
+    TransferOp,
+    TransferOpGraph,
+    TransferOpStatus,
+    TransferType,
+)
+from flexkv.transfer.scheduler import TransferScheduler
+
+
+class FullSweepScheduler:
+    """The pre-change scheduler, kept here so `--baseline` can print a
+    before/after table from one run instead of asking you to check out the
+    previous revision.
+
+    `tests/test_transfer_scheduler_incremental.py` keeps its own copy, where it
+    serves as the equivalence oracle; keep the two in step.
+    """
+
+    def __init__(self) -> None:
+        self._transfer_graphs = {}
+
+    def add_transfer_graph(self, graph) -> None:
+        self._transfer_graphs[graph.graph_id] = graph
+
+    def schedule(self, finished_ops):
+        for op in finished_ops:
+            if op.graph_id in self._transfer_graphs:
+                self._transfer_graphs[op.graph_id].mark_completed(op.op_id)
+
+        next_ops = []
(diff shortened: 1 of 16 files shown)


linhu-nv (approved):

Luis-xu (commented):
LGTM! 现在worker的launch_transfer()都默认返回True，后续需要根据实际传输成功情况来进行True/False的返回，在此基础之上可以进行语义的扩展，包括返回成功page数量等，为其他功能做基础。 @linhu-nv

PR #240, opened 2026-07-28
feat(transfer): chunked cudaHostRegister via FLEXKV_CUDAHOST_CHUNK_SI…

zhuofan1123:
Register HugePage host buffers in segments matching the existing chunked unregister layout. chunk=0 (default) registers the whole buffer in one call; chunk>0 splits registration to avoid one-shot failures on very large buffers, with rollback of already-registered chunks on error.

Document the new env var in the zh/en config references.


peaceforeverCN (approved):
lgtm

linhu-nv (approved):

PR #245, opened 2026-07-30
Interim solution for dpskv4 CPU capacity issues.

Luis-xu:
A interim solution for dpsk v4 cpu capacity.

### DIFF
--- a/flexkv/common/config.py
+++ b/flexkv/common/config.py
@@ -109,7 +109,17 @@ class ModelConfig:
     num_layers: int = 1
     num_kv_heads: int = 1
     head_size: int = 1
+    # Attention-arch flag (DeepSeek MLA vs MHA). Used for head-parameter
+    # derivation and as the default for layout/share axes below.
     use_mla: bool = False
+    # Physical KV layout: True => kv_dim=1 (one packed stream per slot).
+    # None inherits ``use_mla`` (see ``is_layout_mla``).
+    layout_is_mla: Optional[bool] = None
+    # Whether CP ranks share the same *logical KV content* (classic MLA).
+    # Does NOT change process identity: ``effective_tp_rank`` always includes
+    # ``cp_rank``. Used for copy/capacity semantics (e.g. MLA D2H all_write).
+    # None inherits ``use_mla``. DSv4: False + ``layout_is_mla=True``.
+    share_kv_across_cp: Optional[bool] = None
     dtype: torch.dtype = torch.bfloat16
 
     # ------------------------------------------------------------------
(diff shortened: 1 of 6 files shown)


flexkv/integration/config.py
```diff
@@ -422,6 +421,18 @@ def post_init_from_sglang_config(
 
         self.model_config.use_mla = use_mla
 
+        # Approach 1: identity always includes cp_rank; layout/share are separate.
+        # Set DSv4 axes before GB->blocks so kv_dim uses layout_is_mla.
+        is_dsv4 = bool(getattr(sglang_config, "is_deepseek_v4_arch", False))
+        if is_dsv4:
+            self.model_config.layout_is_mla = True
+            self.model_config.share_kv_across_cp = False
```
XingLiu1, line 429: [P1] `share_kv_across_cp=False` currently changes capacity accounting, but it is not propagated to the data-transfer path.

DSv4 registers packed single-stream GPU layouts (`KVCacheLayout.is_mla=True`). `tpGPUCPUTransferWorker` then derives `self.is_mla` from that physical layout and forwards it to `TPTransferThreadGroup` and the layerwise transfer group as the logical-sharing flag. In C++, `is_mla=True` selects shared-KV behavior: the default `sharded` D2H path copies only `gpu_chunk_size / num_gpus` from each rank into one assembled CPU block, while `all_write` places every rank in a separate full-KV region. The transfer group size is `effective_tp_size_per_node = tp_size * cp_size`.

That is not valid for this DSv4 configuration: CP ranks own different logical KV content. In default `sharded` mode, most of each CP rank local KV is discarded. In `all_write`/layerwise mode, the C++ offset scheme still assumes N full per-rank regions, while the new capacity calculation deliberately skips the N-way divisor for `share_kv_across_cp=False`; the data-plane addressing and configured capacity therefore disagree.

Please propagate a separate sharing/copy-semantics flag through the transfer workers and C++ interfaces. `layout_is_mla` should control only `kv_dim` and physical strides. `share_kv_across_cp` (or a more general per-rank-sharing flag) should control sharded/all-write behavior and CPU rank offsets. For `False`, each effective `(cp_rank, tp_rank)` needs its own non-shared CPU slice even though `kv_dim == 1`.

A regression test should cover `tp_size=2, cp_size=2, layout_is_mla=True, share_kv_across_cp=False` for both whole-block and layerwise round trips, including the default `sharded` mode and `all_write`.

flexkv/common/config.py
```diff
@@ -356,6 +366,25 @@ def effective_tp_size_per_node(self) -> int:
...
+
+        Falls back to ``use_mla`` when ``layout_is_mla`` is unset.
+        """
+        return self.use_mla if self.layout_is_mla is None else bool(self.layout_is_mla)
+
+    def is_kv_shared_across_cp(self) -> bool:
+        """Whether CP ranks share the same logical KV content.
+
+        Falls back to ``use_mla`` when ``share_kv_across_cp`` is unset.
+        Does not affect ``effective_tp_rank`` identity (approach 1).
```
XingLiu1, line 380: [P2] The documented “approach 1” identity rule is not implemented by `RankInfo.effective_tp_rank`.

The new field comment says process identity always includes `cp_rank`, and this method says the sharing flag does not affect identity. However, `effective_tp_rank` still branches on `model_config.use_mla` and returns only `tp_rank` whenever `use_mla=True`. Thus the newly expressible combination `use_mla=True, share_kv_across_cp=False` gives every CP rank the same effective rank, even though `effective_tp_size` still counts `tp_size * cp_size`. That can collide in registration/worker/eventfd maps or leave the system waiting for unique ranks that can never appear.

DSv4 currently avoids this particular branch because it reports `use_mla=False`, but the purpose of these new axes is to make architecture, layout, sharing, and identity independent. Please either implement the stated rule unconditionally as `cp_rank * tp_size + tp_rank`, or, if legacy shared MLA must collapse CP identities, derive both rank identity and effective group size consistently from the explicit sharing flag and update the documentation accordingly. A small test enumerating all CP/TP ranks would catch the duplicate-rank case.

PR #253, opened 2026-08-06
docs: clarify SGLang FlexKV version guidance

XingLiu1:
## Summary

- document that the base FlexKV integration is available in upstream SGLang starting from v0.5.16 and no longer requires a patch
- document that DeepSeek V4 support still requires the unmerged SGLang PR #31781, pinned to commit `ee0465a09196421a6e4d53a3103eccdef1dd32ac`
- add a Chinese version of the SGLang integration README
- mark `sglang_flexkv_connector.patch` as a legacy reference that should not be used with supported releases or the DeepSeek V4 path

## Why

The existing README still instructed users to check out the old `deepseek_v4` branch and apply a manually ported patch. That workflow became stale after the base integration was merged upstream, while the separate DeepSeek V4 adaptation has not yet been merged.

## Validation

- `git diff --check`
- verified that the SGLang `v0.5.16` tag contains the base FlexKV integration
- verified that SGLang PR #31781 remains open and its current head is `ee0465a09196421a6e4d53a3103eccdef1dd32ac`

### DIFF
--- a/flexkv/integration/sglang/README.md
+++ b/flexkv/integration/sglang/README.md
@@ -1,101 +1,87 @@
-# SGLang FlexKV Connector Integration
+# Using FlexKV with SGLang
 
-> **WARNING: THIS INTEGRATION IS NOT YET COMPLETE AND HAS NOT BEEN TESTED.**
->
-> The patch was mechanically ported from the development branch and conflicts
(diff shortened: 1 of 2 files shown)


XingLiu1:
Superseded by #251. The SGLang README commit has been added to the CI branch so both changes can be reviewed and merged together.

PR #266, opened 2026-08-21
use insert after for all transfer type

linhu-nv:


flexkv/cache/cache_engine.py
```diff
@@ -739,12 +713,10 @@ def insert(self,
                sequence_meta: SequenceMeta,
                physical_block_ids: np.ndarray,
                num_insert_blocks: int = -1,
-               is_ready: bool = True,
                match_result: Optional[MatchResult] = None) -> Optional[RadixNode]:
```
XingLiu1, line 716: [P1] Migrate the regression tests along with the readiness API removal

`CacheEngine.insert()` no longer accepts `is_ready`, but five test files still contain 76 calls using that keyword, and other tests still assert the removed readiness fields/helpers. The existing CI already stops at `test_cancel_task_rollback.py::test_rollback_removes_unready_leaf_and_recycles[CacheEngine]` with `TypeError: insert() got an unexpected keyword argument 'is_ready'` (`1 failed, 1 skipped`). Please migrate these callers and replace the old unready-node rollback expectations with detached-staging, completion-publication, and cancellation-recycling coverage so the new lifecycle is actually exercised.

flexkv/cache/cache_engine.py
```diff
@@ -2354,135 +2215,71 @@ def _put_impl_global(self,
...
+
+        # Existing cached blocks may feed H2DISK/H2REMOTE while the graph is
+        # running. Pin their deepest node until every consumer finishes.
+        cpu_anchor = cpu_matched_result.last_node
+        if len(cpu_matched_blocks) > 0 and cpu_anchor is not None:
+            node_to_unlock[DeviceType.CPU] = cpu_anchor
+        ssd_anchor = ssd_matched_result.last_node
+        if (put_to_ssd and len(ssd_matched_blocks) > 0
+                and ssd_anchor is not None):
+            node_to_unlock[DeviceType.SSD] = ssd_anchor
```
XingLiu1, line 2272: [P2] Keep the existing remote prefix pinned until deferred PUT commits

This list protects CPU and SSD anchors but omits the ordinary, non-Mooncake remote anchor. `remote_cache_engine.take(..., protected_node=...)` only protects that node during allocation; once planning returns, a concurrent allocation can evict the existing remote prefix before H2REMOTE finishes. `_commit_deferred_insert()` then sees `current_blocks < staged_start_block` and recycles the successfully written suffix.

Reproduced with a four-block remote pool containing a two-block prefix: plan a PUT extending it to four blocks, allocate two remote blocks for an unrelated sequence, then complete the first PUT. The base retains a four-block remote hit; this head has zero, while CPU has all four. A retry can also be skipped by the CPU-full-hit early return. Please add the existing remote anchor to the plan's lock/unlock ownership, including abort cleanup.

flexkv/cache/cache_engine.py
```diff
@@ -2647,57 +2443,55 @@ def _put_impl_local(self,
...
+        deferred_inserts.append(DeferredCacheInsert(
+            device_type=DeviceType.CPU,
+            sequence_meta=sequence_meta,
+            physical_blocks=fragment12_cpu_blocks,
+            staged_start_block=block_mask_start + num_skipped_blocks,
+            remote_start_block=block_mask_start + num_skipped_blocks,
+            requested_end_block=requested_end,
+            swa_slot=cpu_swa_slot,
+            publish_to_peer=self.cache_config.enable_p2p_cpu,
+        ))
```
XingLiu1, line 2461: [P2] Preserve CPU availability at D2H completion when lower-tier writes are pending

Moving CPU publication exclusively to graph completion changes the local CPU+SSD PUT path, but D2H remains in `finished_ops_ids` and `KVTaskManager.check_completed()` still allows success at that task-end. After D2H completes, a subsequent lookup now misses the newly saved CPU data until H2DISK also completes; if H2DISK later fails, `_fail_task()` aborts the whole plan and discards even the valid CPU copy.

The base/head control-plane comparison gives readable CPU hits of 4/0 blocks both immediately after D2H and after a subsequent SSD failure. This makes CPU reuse depend on SSD latency and failures. Please publish each tier after its own transfer completes, retaining the pins needed by remaining consumers, or explicitly reconcile whole-graph publication with the existing early-success PUT contract.

XingLiu1 (changes requested):
Reviewed head `07496038d0f2fd33e02fd635f427d6db1a056b85` against base `34ab141e87763536de27d5aa3b80ae720abaf279`.

Requesting changes for the three inline findings: the readiness API migration leaves CI failing, ordinary remote PUT loses the prefix protection needed for deferred publication, and CPU publication now waits for lower-tier writes while PUT still supports success at D2H completion.

The two runtime findings were checked against both revisions using the actual Python planning, radix, mempool, completion, and abort code, with external dependencies and completion events replaced by a small control-plane harness:

| Scenario | Base | PR head |
| --- | --- | --- |
| Remote hit after extending a 2-block prefix to 4 blocks under concurrent allocation pressure | 4 blocks | 0 blocks |
| Readable CPU hit after D2H completes, while SSD write is pending | 4 blocks | 0 blocks |
| CPU hit retained after the subsequent SSD failure and drained-graph abort | 4 blocks | 0 blocks |

The [existing CI failure](https://github.com/taco-project/FlexKV/actions/runs/32454114727/job/96688003532) was also verified directly. These are control-plane checks, not GPU, native C++ extension, real storage I/O, or hardware performance validation; the full test suite was not rerun locally.

XingLiu1 (approved):

PR #270, opened 2026-08-24
[bugfix] harden worker drain shutdown and EOF handling

Claisenn:
## Summary

Harden the worker batch-drain shutdown path without changing upstream's current
control-message, transfer-result, or tracing behavior.

Upstream already fixed the original first/single-op batching regression. Two
edge cases remain:

1. `multiprocessing.Connection.poll()` reports a peer-closed pipe as readable.
   If `recv()` raises EOF while draining, the already-received batch was dropped
   before its completions were reported.
2. A `None` sentinel drained after real ops called subclass `shutdown()` inside
   `run()`, then `_worker_process.finally` called it again. Some subclasses run
   external unregister work before the base idempotence guard, so cleanup could
   execute twice.

## Fix

- Catch EOF inside the drain loop, process the accumulated batch, then return.
- Treat a drained sentinel the same way: process prior ops and let
  `_worker_process` own the single shutdown call.
- Preserve upstream dictionary control ops, trace metrics, and
  `WorkerTransferResult` handling unchanged.

## Verification

- Focused regression tests drive the real `TransferWorkerBase.run` with fake
  pipes and upstream-style result tuples.
- AST execution verifies already-received ops complete across drain EOF.
- Python compile, ruff, and `git diff --check` pass for newly added code; the
  remaining worker.py lint findings are identical to upstream main.

### DIFF
--- a/flexkv/transfer/worker.py
+++ b/flexkv/transfer/worker.py
@@ -538,11 +538,17 @@ class TransferWorkerBase(ABC):
                 # continue would otherwise drop the first op forever (which
                 # stalls D2H → H2REMOTE and leaves mooncake PutStart=0).
                 batch_ops = [op]
-                shutdown_after_batch = False
+                stop_after_batch = False
                 while self.transfer_conn.poll(timeout=0):
-                    op = self.transfer_conn.recv()
+                    try:
+                        op = self.transfer_conn.recv()
+                    except EOFError:
+                        # A closed pipe is readable. Preserve the batch already
+                        # received, then exit after reporting its completions.
+                        stop_after_batch = True
+                        break
                     if op is None:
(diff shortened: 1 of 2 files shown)


Claisenn:
Review request: @linhu-nv @Luis-xu — this is the minimal upstream-main hardening for worker drain EOF and single-owner shutdown behavior.

Claisenn:
CI note: the wheel builds successfully; the CPU suite fails in the pre-existing `test_global_prefetch_ready` fixture because `_cache_tree_lock` is missing. This is exactly fixed by #268, whose CI is green. No code from this PR is on that failing path.

linhu-nv (approved):

PR #271, opened 2026-08-24
[feature] add StreamingLLM sink-aware eviction

Claisenn:
## Summary

Add complete StreamingLLM attention-sink protection to FlexKV's radix-tree
external cache eviction, including correctness fixes and O(1) classification.

This PR intentionally keeps the feature, fallback correctness, and its hot-path
optimization together as four reviewable commits:

1. protect the first configured number of logical blocks;
2. guarantee allocation progress when all remaining candidates are sinks;
3. preserve protection when cascade eviction exposes a sink parent, with
   dynamic normal/sink heaps;
4. cache each node's root block offset so sink classification is O(1).

## Configuration

```bash
export FLEXKV_SINK_BLOCK_COUNT=N
```

`N=0` (default) preserves existing behavior.

## Correctness

- Non-sink candidates are always exhausted before fallback considers sinks.
- Fallback is re-evaluated after every structural mutation, so a newly exposed
  sink parent cannot enter the normal heap and allocation cannot stop early.
- Fallback orders sink candidates by the configured policy instead of assigning
  every sink infinite priority.
- Python and C++ implementations preserve identical split/insert/merge offset
  invariants; distributed bulk rebuild runs one off-hot-path offset pass.

## Performance

Before: sink classification walked ancestors for every eviction candidate,
`O(candidates * tree_depth)` on the serialized control plane.

After: each node stores its block offset, maintained in O(1) on insert/split;
classification is a single integer read. A CPU lookup-only microbenchmark at
`depth=256`, `4096` candidates measured ~81.7x speedup for this predicate
(not an end-to-end eviction claim).

## Verification

- Deterministic dynamic-fallback tests: all non-sinks precede sink parents, and
  all-sink fallback follows raw policy priority.
- Randomized real-Python radix differential test: 229,404 cached-offset checks,
  zero failures.
- Standalone C++ offset model: 24,000 randomized mutation checks, zero failures.
- C++ radix/local syntax checks pass with CPU libtorch after adapting only two
  pre-existing Torch 2.13 API differences in temporary copies.
- Combined integration rehearsal with worker fix, vLLM layer-wise loading, and
  decayed LFU passes all available CPU/static tests.

### DIFF
--- a/csrc/bindings.cpp
+++ b/csrc/bindings.cpp
@@ -930,16 +930,17 @@ PYBIND11_MODULE(c_ext, m) {
   py::class_<flexkv::CRadixTreeIndex>(m, "CRadixTreeIndex")
       .def(py::init([](int tokens_per_block, unsigned int max_num_blocks,
                        int hit_reward_seconds, std::string eviction_policy,
-                       int protected_threshold) {
+                       int protected_threshold, int sink_block_count) {
              auto policy = flexkv::parse_eviction_policy(eviction_policy);
              return new flexkv::CRadixTreeIndex(
                  tokens_per_block, max_num_blocks, hit_reward_seconds, policy,
-                 protected_threshold);
+                 protected_threshold, sink_block_count);
            }),
            py::arg("tokens_per_block"), py::arg("max_num_blocks") = 1000000,
            py::arg("hit_reward_seconds") = 0,
            py::arg("eviction_policy") = "lru",
-           py::arg("protected_threshold") = 2)
+           py::arg("protected_threshold") = 2,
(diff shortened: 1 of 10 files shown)


Claisenn:
Review request: @linhu-nv — this PR consolidates StreamingLLM sink protection, dynamic fallback correctness, and O(1) cached sink offsets.

Claisenn:
CI note: if the CPU suite reports `test_global_prefetch_ready` / missing `_cache_tree_lock`, that is the known upstream fixture issue fixed by #268 (green), unrelated to this sink-eviction diff.

PR #272, opened 2026-08-24
[feature] pipeline FlexKV loads layer by layer in vLLM

Claisenn:
## Summary

Wire FlexKV's existing layer-wise transfer/eventfd data plane into the vLLM V1
connector so external KV GETs overlap model execution layer by layer.

Before this change, FlexKV returned `is_async=True`; vLLM put the request in
`WAITING_FOR_REMOTE_KVS` and ran no forward pass until the entire transfer
finished. The worker-side `start_load_kv()` and `wait_for_layer_load()` hooks
were no-ops, so the existing LayerwiseTransferWorker could not provide overlap.

## Design

- Scheduler launches all GETs in the step as one fused LAYERWISE graph.
- A triple-buffered `counter_id` is included in concrete scheduler metadata.
- Worker sends per-layer eventfds to LayerwiseTransferWorker before cache
  registration can deadlock waiting for the handshake.
- `start_load_kv()` binds the step counter; each attention layer consumes one
  eventfd in `wait_for_layer_load(layer_name)`.
- Layerwise matches return `is_async=False`, allowing the request to enter the
  same forward step instead of WAITING_FOR_REMOTE_KVS.

Expected timing changes from roughly `T_transfer + T_forward` to first-layer
startup plus `sum(max(T_transfer_layer, T_compute_layer))`.

## Safety

- Enabled only by `kv_connector_extra_config.use_layerwise=true`; env-only
  enablement is ignored because vLLM must select PIECEWISE CUDA graph mode.
- Context parallelism and GDS fail fast until their rank/graph contracts are
  implemented for this path.
- Per-layer waits have a configurable timeout and poison the counter pool after
  failure, preventing stale eventfd tokens from satisfying a later step.
- A new step cannot bind while the previous counter is active.
- Short requests that finish before the GET graph completion tail retain their
  blocks and defer PUT creation until the GET response is observed.
- Failed deferred GETs release blocks without issuing a PUT.

## vLLM wrapper change

`examples/vllm_adaption/vllm_main-flexkv-layerwise-wrapper.patch` adds:

- `requires_piecewise_for_cudagraph()` for `use_layerwise`;
- metadata bind/clear delegation into the installed FlexKV implementation;
- upstream unit coverage for both behaviors.

This small change should be submitted to vLLM separately.

## Verification

- 9 CPU-only tests pass:
  - counter round-robin and launch policy;
  - exactly-once per-layer waits;
  - no-load behavior and invalid metadata;
  - stale-step rejection and timeout poisoning;
  - fd cleanup;
  - real SCM_RIGHTS wire protocol;
  - NACK/reconnect/ACK retry.
- AST execution of the real adapter methods verifies:
  - layerwise hit returns `is_async=False`;
  - launch uses `as_batch=True`, `layerwise_transfer=True`, matching counter id;
  - deferred GET completion creates PUT and keeps blocks pinned;
  - failed deferred GET releases blocks without PUT.
- vLLM wrapper patch applies cleanly to current upstream source and compiles.
- Independent targeted review: APPROVE after four high-risk findings were fixed.

GPU end-to-end validation is still required for TTFT measurement and CUDA
layer/eventfd timing; existing LayerwiseTransferGroup GPU tests cover the data
plane independently.

### DIFF
--- a/flexkv/integration/vllm/layerwise_sync.py
+++ b/flexkv/integration/vllm/layerwise_sync.py
@@ -37,6 +37,18 @@ def _linux_eventfd(initval: int = 0, flags: int = _EFD_SEMAPHORE) -> int:
     return int(fd)
 
 
+def _settle_eventfd(fd: int, timeout_s: float) -> bool:
+    """Consume one pending eventfd unit, waiting at most timeout_s."""
+    try:
+        ready, _, _ = select.select([fd], [], [], timeout_s)
+        if not ready:
+            return False
+        os.read(fd, 8)
+        return True
+    except OSError:
+        return True
+
+
 def _read_eventfd(fd: int) -> int:
     timeout_s = float(os.getenv("FLEXKV_LAYERWISE_WAIT_TIMEOUT_S", "60"))
     ready, _, _ = select.select([fd], [], [], timeout_s)
@@ -97,11 +109,30 @@ class LayerwiseStepCoordinator:
         return bool(needs_load and not self.enabled)
 
     def launch_kwargs(self, metadata: LayerwiseLoadMetadata) -> dict[str, object]:
+        # A no-load step carries counter_id=-1. Clamping that to 0 would make
+        # the data plane signal counter 0 -- which a concurrent real load may
+        # own -- so its layers could see units that belong to no transfer.
+        # Only a metadata that actually has a load may drive a layer-wise
+        # launch; otherwise fall back to the non-layer-wise path.
+        if not metadata.enabled or not metadata.has_load:
+            return {
+                "as_batch": metadata.enabled,
+                "layerwise_transfer": False,
+                "counter_id": 0,
+            }
+        self._validate_launch_counter(metadata.counter_id)
         return {
             "as_batch": metadata.enabled,
             "layerwise_transfer": metadata.enabled,
-            "counter_id": max(metadata.counter_id, 0),
+            "counter_id": metadata.counter_id,
         }
+
+    def _validate_launch_counter(self, counter_id: int) -> None:
+        if counter_id < 0 or counter_id >= self.num_counters:
+            raise ValueError(
+                f"layer-wise launch counter_id={counter_id} outside "
+                f"[0, {self.num_counters})"
+            )
 
 
 class LayerwiseCounterPool:
@@ -120,6 +151,7 @@ class LayerwiseCounterPool:
         fd_factory: Callable[[], int] = _linux_eventfd,
         fd_reader: Callable[[int], int] = _read_eventfd,
         fd_closer: Callable[[int], None] = os.close,
+        fd_settler: Callable[[int, float], bool] = _settle_eventfd,
     ) -> None:
         if not layer_names:
             raise ValueError("layer_names must not be empty")
@@ -136,6 +168,10 @@ class LayerwiseCounterPool:
         self.num_counters = num_counters
         self._fd_reader = fd_reader
         self._fd_closer = fd_closer
+        # Retained for compatibility with external test fixtures that inject a
+        # settler. Reusing an incomplete asynchronous counter is unsafe, so
+        # this pool deliberately no longer invokes it.
+        self._fd_settler = fd_settler
         self._fds = [
             [fd_factory() for _ in range(self.num_layers)]
             for _ in range(num_counters)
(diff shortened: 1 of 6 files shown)


Claisenn:
Review request: @peaceforeverCN @linhu-nv — this adds the vLLM layer-wise metadata/eventfd path and PIECEWISE CUDA-graph contract.

Claisenn:
CI note: if the wheel CPU suite reports `test_global_prefetch_ready` / missing `_cache_tree_lock`, that is the known upstream fixture issue fixed by #268 (green). vLLM compatibility checks for this PR are already green.

Claisenn:
Hi @linhu-nv, this PR is ready for review. All current CI checks are green, including the wheel/CPU test suite and the vLLM compatibility checks on Python 3.10 and 3.11.

Could you please focus on these three areas when convenient?
1. the layer-wise metadata/eventfd handshake and stale-step protection;
2. deferred PUT creation and KV-block lifetime handling;
3. the explicit PIECEWISE CUDA graph requirement and CP/GDS fail-fast guards.

The feature is opt-in through `kv_connector_extra_config.use_layerwise`, so the existing non-layer-wise path remains unchanged by default. The only remaining environment-dependent validation is an end-to-end GPU run; feedback or help validating that path would be appreciated.

Since I am an external contributor and cannot create a formal review request for repository maintainers, an approval or review feedback would help unblock the PR. Thank you!

linhu-nv:
Hi @Claisenn, thanks a lot for the great contribution, actually this is what exactly we need. But I am on a business trip this week, and sorry for the delayed review. I will tag some maintainers for your pr, and get back to them later this week. Thanks again!

flexkv/integration/vllm/vllm_v1_adapter.py
```diff
@@ -1030,6 +1229,14 @@ def reset_cache(self) -> Optional[bool]:
             return self.connector.reset_cache()
         return None
 
+    def bind_connector_metadata(
+        self, metadata: FlexKVConnectorMetadata
+    ) -> None:
+        self._bound_metadata = metadata
+
+    def clear_connector_metadata(self) -> None:
+        self._bound_metadata = None
```
linhu-nv, line 1361: 这里不释放 counter pool，所以任何一次 forward 中途异常，包括oom，超时，提前abort都会让 `_active_counter` 永久卡住，下一步 `bind()` 抛->引擎死。

  这是 step 的自然边界，应该无条件收尾。建议配合 pool 侧加一个退役接口：

  ```python
  def retire_counter(self, counter_id: int) -> None:
      """A counter whose step exited abnormally may hold stale eventfd counts.
      Never reuse it; the transfer worker may still write to those fds."""
      self._retired.add(counter_id)
      self._active_counter = -1
 ```
  - 这里无条件调一次收尾：正常走完 release()，否则 retire_counter()
  - 轮转时跳过 retired，num_counters 从写死的 3 提到可配、默认 8
  - 全部退役后不再崩，而是永久回落同步路径（external_match_is_async 恢复返回 needs_load）+ 一条 error 日志

  不用 drain：超时后传输仍可能在飞，drain 完再被写一次还是脏；重建 fd 又需要重传，而 UDS 通道在 _finish_layerwise_eventfd_sender() 已经关了。退役是最简单的安全解。

flexkv/integration/vllm/layerwise_sync.py
```diff
@@ -0,0 +1,267 @@
...
+                f"{self.layer_names!r}"
+            ) from exc
+
+        with self._lock:
+            if self._waited[counter_id][layer_index]:
+                return
+        try:
+            self._fd_reader(self._fds[counter_id][layer_index])
+        except BaseException as exc:
+            self._failed_error = exc
```
linhu-nv, line 237: 这里改成 `self.retire_counter(counter_id)` + 继续抛（那个请求确实废了，交给 vLLM 在请求级处理），`_failed_error` 这个字段就可以整个删掉

linhu-nv:
我这边也在进行一些测试的工作，这个pr确实对flexkv比较重要，所以我想也同步验证一下，麻烦耐心等待，谢谢

linhu-nv:
can you please do the rebase? @Claisenn

Claisenn:
Hi @linhu-nv, done — I rebased this PR onto the latest `main` (`016c290`).

I resolved the two conflicts by preserving both sides of the updated contracts:
- kept the new `kv_cache_config` forwarding from `main` together with the layer-wise connector setup;
- kept the newer keyword-only `LayerwiseTransferGroup` initialization from #277/#278, including `ce_force_path=-1`.

I also verified the old and rebased histories with `git range-diff`: both original PR commits are still present as two corresponding rebased commits (no squash/drop). The layer-wise CPU regression suite passes locally (`12 passed, 1 skipped`), and the PR is now reported as mergeable. CI is running again on the rebased head. Thanks!

linhu-nv (approved):

linhu-nv:
Merged. thanks again for the great work

PR #276, opened 2026-08-27
refactor(transfer): pool identity + completion contracts; retire the layerwise worker

linhu-nv:
## What this is

The transfer-layer refactor from `docs/transfer_runtime_refactor_design_zh.md`,
without `RuntimeCore`. Two ideas drive nearly all of it:

1. **A pool is a slot-id space, not a kind of transfer.** Full KV and SWA share
   direction, tiers, layout family and worker, so `is_swa: bool` forced every
   consumer to re-derive the pool (`_swa_worker_map` beside `_worker_map`,
   `swa_cpu_blocks` beside `cpu_blocks`). `PoolId` makes it a value on the op
   and a key in the dispatch registry.

2. **Per-layer notification is a completion contract, not a worker class.**
   `LayerwiseTransferWorker` plus its 1684-line C++ twin duplicated an entire
   stack to vary *one thing*: when completion is reported. It becomes
   `CompletionContract.PER_LAYER` on the ordinary GPU<->CPU worker.

Net effect: **-9.8k lines deleted**, `worker.py` goes from 4060 lines to a
package, and `csrc/layerwise.{cpp,h}` (2048 lines) is gone.

## Intended PR split

| # | Commit | Why separable |
|---|--------|---------------|
| 1 | `PoolId, not is_swa` | Mechanical; foundation for everything after |
| 2 | `split worker.py into workers/` | Pure moves; review as "moved, unchanged" |
| 3 | `per-layer completion as a contract` | **Largest / needs the most care** |
| 4 | `region batching + NIXL as a backend` | Independent seam |
| 5 | `rewire engine onto the registry` | Carries the 2 concurrency fixes |
| 6 | `split resident H2D` | **Unrelated perf fix** -- lands alone |

Commit 6 is genuinely independent (GET planning only): a mixed-tier hit used to
make CPU-resident blocks wait on the SSD/remote reads; now the resident lane
launches immediately and overlaps them.

## Two concurrency fixes (in commit 5)

* `WorkerHandle.shutdown()` is idempotent and holds its lock for the whole body
  -- `terminate()`/`close()` from a second caller raced the first caller's
  `join()`/`close()`, and `__del__` could fire it a third time from the GC
  thread.
* `_collect_worker_handles()` dedupes **by object identity**. Several
  `(pool, type)` keys map to one worker, and the caller starts one shutdown
  thread *per element* -- duplicates meant several threads racing on one
  process.

## Validation

e2e, all inside containers on 8xH20:

| Model | Result |
|---|---|
| Qwen3-8B | **bit-exact** vs baseline; perf parity |
| GLM-5.2-FP8 (MLA) | e2e green |
| DeepSeek-V4-Flash | needle-in-haystack **72/72**, both transfer modes, at 98.98% hit rate |

DSv4 was run at a verified hit rate deliberately -- a correctness claim with
`cached_tokens=0` proves nothing.

## Known issue, pre-existing, *not* introduced here

On DSv4 the tail logprobs right after a restore boundary deviate by ~11 nats
against a 4e-07 noise floor. Root cause: FlexKV assumes KV is content-addressed
and portable across requests, but DSv4's SWA ring is keyed by `state_slot`/`pos`
-- per-request state. A restore trusts one page of ring written by *whichever*
request stored it.

I confirmed this is not a regression from this branch: the unrefactored baseline
(`c213e5d`) **cannot run DSv4 at all** (`TypeError: Argument 'token_ids' has
incorrect type (expected list, got array.array)`, all TP ranks), and the
one-page contract is identical on both sides. DSv4 is newly onboarded, so the
mismatch is newly *exposed*, not newly *introduced*.

Impact is precision, not semantics -- needle is 72/72 -- so it affects exact
logprob consumers (RL rewards, perplexity, spec-decode acceptance). Fixing it
belongs in a separate PR; capping the match key is **not** the fix (it turns the
lookup into a strict prefix, which DSv4's snapshot-depth rule rejects wholesale,
dropping the hit rate to 0).


staryxchen (commented):
## The two design calls
Agree with both:
1. A pool is a slot-id space, not a kind of transfer. `PoolId` on the op and in the dispatch key is the right replacement for `is_swa` plus a parallel worker map.
2. Per-layer notification is a completion contract, not a worker class. `CompletionContract.PER_LAYER` on the ordinary GPU↔CPU worker is the right replacement for `LayerwiseTransferWorker` / `csrc/layerwise.*`.
Those two are the load-bearing changes. They do not imply RuntimeCore, a Plan IR, or two scheduling lanes.
## Plan vs what this branch actually needs
The proposal (`docs/transfer_runtime_refactor_proposal_zh.md`) describes a target that this branch correctly does **not** implement: PlanTemplate/PlanInstance, RuntimeCore, dual lanes, C++ continuation, prefetch generation. Please do not merge that document as approved architecture. As written it will be read as a license for the next PR to grow an interpreter around a DAG we do not have.
`TransferTemplate` is a different object and can stay. It answers *where the bytes sit* (per-group strides, offsets, `layer_milestones`). Plan would answer *what runs in what order* (copy / join / signal / finalize, lanes, leases). Today's intra-op control flow is still one nested loop plus a completion contract; that does not need an IR. Geometry being computed three times in three workers *is* a real problem, and a pure layout table is the minimal fix.
If the geometry helper stays, please keep the name honest (`compile_geometry` / request reuse, not `plan`). `test_layerwise_plan_reuse.py` is testing RegionRequest reuse, which is worth keeping under a name that does not say Plan.
**Lost capability if we drop Plan: none of this PR's runtime behavior.** Plan is not in the tree. Relative to the proposal's end state we would not have a model-varying intra-op DAG, a cached execution graph, or C++ SSD→H2D continuation. Per-layer eventfd, multi-pool, unfused DISK2H→H2D, and shared geometry do not depend on it. Those last two items are not justified by current measurements.
## How to land it
Please do not merge this branch as one PR. Suggested split, close to your commit table but with 4 and 6 taken off the layerwise path:
**PR-A — `PoolId` only.** Mechanical. `is_swa` remains a derived property. No file split, no layerwise deletion, no backend changes.
**PR-B — engine dispatch on `(PoolId, TransferType)`.** Drop `_worker_map` / `_swa_worker_map`. The two shutdown fixes (idempotent `WorkerHandle.shutdown`, dedupe-by-identity in `_collect_worker_handles`) belong here with their regression tests. Split `worker.py` → `workers/` as a *pure move* in its own commit/PR, not mixed with B's behavior change.
**PR-C — the one large PR: contract + unfuse layerwise.** Delete `LayerwiseTransferWorker` and `csrc/layerwise.{cpp,h}`. `PER_LAYER` on the ordinary GPU↔CPU worker. Eventfd handshake stays (`layer_eventfd` / `layer_notify`). `TransferType.LAYERWISE` / `LayerwiseTransferOp` leave the graph: DISK2H is a normal predecessor, H2D carries the contract. `RegionBatchGroup` only to the extent `submit_layerwise` needs it. `TransferTemplate` (geometry + milestones) can come in here. Gate: existing sglang import path for the eventfd helper still resolves; layerwise eventfd timing + empty-layer cases (empty layer is a hang, e2e will not attribute it); at least one real layerwise e2e, not needle-only.
**Keep out of PR-C** (these are other seams):
- NIXL-as-backend, `EdgeGeometry`, `StorageBackend.attach` signature change, `DeviceThreadPool`
- Mooncake MR-split / partial-completion ports — land on main first if they still apply
- Resident H2D overlap (`gpu_bind_offset`) — you already called this independent; it is GET planning, not the worker
- nvcomp `kv_dim` bool, PCFS BLOCKFIRST, Cython `[-1]` under `wraparound=False`, tp==1 MHA guard — correctness bugs, own PRs to main
- `test: drop the cases e2e reaches anyway` — please revert. A refactor is exactly when pickle dtype, double shutdown, empty-layer hang, and `gpu_bind_offset` need to stay pinned. e2e does not replace them.

**PR-D — NIXL / Mooncake as `StorageBackend`.** Own architecture line; RDMA/storage risk, not layerwise correctness.
**PR-E — resident H2D overlapping SSD/remote.** Own PR, own test for `gpu_bind_offset`.
## Freeze for this series
1. `CompletionContract` stays `{WHOLE, PER_LAYER}`. No nodes, lanes, or PlanTemplate.
2. `TransferTemplate` stays geometry. No copy/join/signal DAG.
3. Workers stay one-per-resource-edge, still processes. No RuntimeCore / owner thread in this series. `DeviceThreadPool` rides with PR-D if NIXL needs it, not with PR-C.

linhu-nv:
close as we have separated prs now.

PR #288, opened 2026-09-07
fix(nvcomp): two guards that reject valid transfers

linhu-nv:
2 个独立 bug，都在压缩路径上，都会让**合法**的传输被拒：

1. `csrc/compression/ans/nvcomp_ans_tp.cpp` — size-table guard 把每个`tp==1` 的 MHA 传输都判成非法。
2. `csrc/compression/ans/ans_bindings.cpp` — pybind 边界把 `kv_dim` 收缩成 bool，`kv_dim=1`（MLA）和 `kv_dim=2`（MHA）无法区分

### DIFF
--- a/csrc/compression/ans/ans_bindings.cpp
+++ b/csrc/compression/ans/ans_bindings.cpp
@@ -32,7 +32,7 @@ static size_t transfer_kv_blocks_ans_binding(
     int64_t chunk_size_in_bytes,
     int start_layer_id,
     int num_layers,
-    bool kv_dim,
+    int kv_dim,
     int gpu_block_type,
     int64_t cpu_size_table_ptr,
     int64_t cpu_size_table_block_stride,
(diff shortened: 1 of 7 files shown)


staryxchen (commented):
The two C++ fixes are right, and this is the correct extract from #276.

`kv_dim: bool → int` is the one that actually unblocks tp==1 MHA on *this* base: tp==1 goes through `GPUCPUTransferWorker` / `transfer_kv_blocks_ans_comp`, not `tp_group_transfer_ans`. The size-table guard change matches the kernel (`ptr + i * rank_stride` only moves when `num_gpus_ > 1`) and the allocator (`canonical` when `tp_size==1`). The e2e story in the PR body is the #276 merged worker; on `main` that guard is not on the tp==1 production path.

Please also take the production `num_kv_heads` pass that #257 made necessary and that #276 already has. `NvcompGpuCpuTpStrategy._dispatch` still calls `tp_group_transfer_ans` positionally without it (`ans_strategy.py` ~193–209). After `kv_dim` the next positional is `_table_ptr`, so the four table arguments shift and the binding is one short — TypeError, which is why the tp unit tests have been dead since c213e5d. This PR fixes the tests; the live TP+nvcomp path on `main` is still that call. Keyword args as in #276:

```python
kv_dim=worker.kv_dim,
num_kv_heads=worker.num_kv_heads,
cpu_size_table_tp_ptr=self._table_ptr,
...

staryxchen (approved):
LGTM

PR #289, opened 2026-09-07
fix(cython): stop indexing lists from the end under wraparound=False

linhu-nv:
Release builds cythonize `flexkv/**/*.py` with `wraparound: False` (setup.py:443-470). Under that directive a negative index on a real list is NOT folded to `len + i` -- Cython emits a raw `PyList_GET_ITEM(o, i)`, which reads off the front of the object. The pure-Python path is unaffected, so every green test run misses this; it only ever fires in a compiled build, where it reads garbage or segfaults.

Three genuine list subscripts:

  transfer.py:597      first[:-1] / last[-1] on .tolist() results
  config.py:743        layers_range[-1] (pp_layers returns a list)
  kvtask.py:210        transfer_handles[-1] right after .append()

Left alone: the numpy/torch subscripts at cache_engine.py:3441, kvtask.py:574 and connector.py:1992 fall back to PyObject_GetItem, for which wraparound is irrelevant, and the `[-1] * n` list literals in mooncake_store_utils.py:362 / layer_eventfd.py:238 are not subscripts.

### DIFF
--- a/flexkv/common/transfer.py
+++ b/flexkv/common/transfer.py
@@ -530,15 +530,15 @@ class TransferOpGraph:
                     else:
                         first = block_ids[:max_show].tolist()
                         last = block_ids[-max_show:].tolist()
-                        # Explicit indices, not ``[:-1]``/``[-1]``: release
-                        # builds cythonize this module with wraparound=False,
-                        # under which negative indices on a real list are not
-                        # folded to len + i.  The numpy slices above are safe
-                        # (they go through PyObject_GetItem); ``first``/``last``
-                        # are lists and are not.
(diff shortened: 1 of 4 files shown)


staryxchen (commented):
The three sites in this PR are real, but they are the colder ones (debug `format_blocks`, `pp_size > 1`, `nnodes > 1`). But the same bug still exists in `worker.py`.

Same shape as the kvtask change is enough:
```python
last = regions[len(regions) - 1]
if last[0] + last[1] != base_ptr + mapped_size:
    ...
```

staryxchen (approved):
LGTM

PR #290, opened 2026-09-07
refactor(transfer): identify a pool by PoolId and dispatch on (pool, type)

linhu-nv:
| commit | 作用 |
|---|---|
| `refactor(transfer): identify a transfer's pool by PoolId, not is_swa` | `is_swa: bool` → `pool_id: PoolId`。bool 只能表达两个池，加第三个池就得再加一个 bool。 |
| `refactor(transfer): split worker.py into one module per resource edge` | 4k 行的 `worker.py` 拆成 `workers/` 包，一条物理边一个模块。**纯移动**：每个符号与移动前逐字节相同，`worker.py` 保留为 façade 继续 re-export，所以两条 import 路径都还有效 |
| `refactor(transfer): dispatch on (pool, transfer type), not on is_swa` | engine 侧的 worker 查找改成 `(PoolId, TransferType)` 二元 key |

**review 重点**：`PoolId` 的引入是否覆盖了所有 `is_swa` 读点（`storage_engine.py`
上保留了 `is_swa` 兼容别名，到 PR 3 的 engine rewire 才摘掉）。


staryxchen (commented):
This is the right *first* slice from #276 (PoolId + `(pool, type)` registry + splitting `worker.py`), but the branch currently lands three other behavioral changes that should not ride with it.

**Keep:** `PoolId` with `is_swa` as a derived alias (conflict = error), the live `_worker_map` / `_swa_worker_map` views, idempotent `WorkerHandle.shutdown`, and identity-deduped `_collect_worker_handles`. `test_pool_identity.py` pins the right properties.

**Please take out, or land as their own PRs:**

1. **Layerwise DISK2H hoist.** Dropping `*_disk2h` from `LayerwiseTransferOp` / `WorkerLayerwiseTransferOp`, always passing `None` into cpp, and turning the fused Step 0a into a predecessor DISK2H on the graph (callbacks at CPU-ready) is the unfuse we called PR-C. It is a real GET semantic change. The C++ Step 0a path is still there. Needs layerwise e2e, not a "mechanical PoolId" review.

2. **`gpu_bind_offset` / `src_is_staged` / `_split_h2d_lanes`.** That is the resident-H2D split. This PR does not change `cache_engine.py`, so nothing sets the flags — but the merge path is already forked. It does not belong here.

3. **WorkerTransferOp pickle encoding** (`_encode_array` / `_encode_ttype`). Unrelated dispatch-thread micro-opt.

**Dispatch fallback should not exist on this base.** `_worker_entry_for` falling back to FULL_KV when a pool has no worker for that type used to be `ValueError`. A SWA DISK2H/REMOTE op hitting the full-KV worker is the wrong slot-id space. This branch still registers *separate* SWA GPU workers — it does not make SWA "a pool on the same GPU↔CPU worker" — so the fallback is not implementing that design; it is a new silent misroute. Misses should still raise.

Nits: the PR body says `tpGPUCPUTransferWorker` was merged into `GPUCPUTransferWorker`; both classes are still in `workers/gpu_cpu.py`. The `NixlFileBackend` string is from a later PR. Several new modules still carry `worker.py`'s import soup (`handle.py` especially); `transfer.py` imports `GLOBAL_CONFIG_FROM_ENV` without using it.

Done for this PR looks like: PoolId on the op, one registry, files split, old error on unknown `(pool, type)`. Unfuse and resident lanes stay out.

staryxchen (commented):
The remaining blocker is still the Layerwise DISK2H hoist. That is the unfuse we called PR-C, not PoolId.

`LayerwiseTransferOp` / `WorkerLayerwiseTransferOp` still drop `*_disk2h`; `merge_to_batch_graph` still lifts DISK2H onto its own graph op (callbacks at CPU-ready) and layerwise only carries H2D; `layerwise.py` still passes `None` into cpp; `test_merge_to_batch_graph_swa_callbacks` still expects 3 ops and fire order `main_disk2h, swa_disk2h, main_h2d, swa_h2d`. GET publish timing and graph shape change; C++ Step 0a is left idle.

It is also inconsistent with the registry this PR just made strict. The merge comments say the hoisted DISK2H is dispatched to the CPU↔SSD worker, but `_init_workers` still skips FULL_KV `DISK2H` when layerwise is on, and the startup assert still requires `DISK2H not in self._worker_map`. After the fallback removal that miss is `ValueError`. SWA `DISK2H` *is* registered in the same function, so the two pools disagree. Revert the hoist (fields, merge, `layerwise.py`, the test) rather than completing it here.

Nits: the comment above `_workers` (~1054) still describes the FULL_KV fallback that `_worker_entry_for` no longer does. The PR body still says `tpGPUCPUTransferWorker` was merged into `GPUCPUTransferWorker`; both classes are still in `workers/gpu_cpu.py`.

linhu-nv:
@staryxchen Thanks — all points are valid. Pushed as a force-update of the stack (pr1 e8cca51, pr2 d003e6c, pr3 7e285d5, pr4 7bd073d).
     
Resident-H2D lanes in the wrong PR — fixed. gpu_bind_offset, src_is_staged / _split_h2d_lanes, and the merged_resident_h2d_op sink now live in the resident-H2D PR that consumes them, and that commit message owns why a GET-planning change touches common/transfer.py.
  
Hand-rolled pickle encoding on WorkerTransferOp — fixed, removed. Two tests were pinned to the encoding rather than the behaviour; rewritten to assert round_tripped.pool_id is pid (identity, not ==: PoolId is an IntEnum, so == PoolId.SWA would also pass for a bare 1 that never interned, and the worker's self._pools[op.pool_id] lookup is what would then fail).
  
Dispatch fallback — fixed, and it was worse than a missing guard. On a (SWA, type) miss the fallback returned the FULL_KV entry; SWA block ids index a different slot space, so the op wrote the right bytes into the wrong pool without raising. _worker_entry_for now raises ValueError with no cross-pool fallback, and the case the fallback was covering — GPU<->CPU, where SWA genuinely rides the same worker — is expressed at registration by registering the same handle under both pools. Sharing a worker is a registration fact, not something dispatch should rediscover by failing a lookup. Covered by test_dispatch_never_crosses_pools and test_dispatch_finds_a_worker_shared_by_both_pools. Also applied to the storage-backends PR, which had carried the fallback forward.
  
Nits (GLOBAL_CONFIG_FROM_ENV, the two NixlFileBackend error strings, handle.py imports) — fixed.

Splitting the layerwise DISK2H hoist into its own PR — I'd rather not, for three reasons:
  
  - It's introduced in the base commit of the stack, and the fields it adds are read by the layerwise plan in the PR above it. Removing it rewrites all four branches, so every diff you've already reviewed changes underneath you.
  - The split doesn't buy reviewability. It's Python-only and self-contained (common/transfer.py's LayerwiseTransferOp fields + merge_to_batch_graph, and transfer/worker_op.py) — transfer/layerwise.py already forwards the main-KV disk2h ids through the existing ssd_block_ids parameter exactly as upstream does, so there's no C++ signature change and nothing to rebuild. It reads as one hunk in place.
  - It costs another full re-stack plus a full suite re-run on all four branches to land the same code in a different order.

linhu-nv:
You're right, and I reproduced it — it's a real bug, not a layering objection.                                                                                                                                                                                             
                                                                                                                                                                                                                                                                  I will revert it — fields, merge, layerwise.py, and the test — rather than completing it here. Completing it means registering a FULL_KV DISK2H worker in layerwise mode and dropping the startup assert, which is a real GET-path change needing layerwise e2e with SSD on; that's PR-C's job. Also fixing the stale _workers comment. Will confirm here once the suite is green.

zhuofan1123 (approved):

PR #292, opened 2026-09-09
[bugfix] Bind CUDA shutdown drains to all owned devices

XingLiu1:
Shutdown launched a fresh daemon thread and called `torch.cuda.synchronize()` without selecting a device. A worker using GPU 3 could therefore synchronize GPU 0 and then unregister host buffers without having drained its own GPU. A single implicit device also misses the other devices used by TP workers.

Both worker and TransferEngine cleanup now use a shared, bounded drain that explicitly binds one daemon to each owned CUDA device. Workers track devices before IPC import, including partial initialization, layerwise/SWA groups and remapping; TransferEngine uses the registered main/SWA GPU IDs for its own process. Each device can finish independently if another hangs. Failed or timed-out worker drains retain registered backing tensors until process exit instead of explicitly unregistering them while DMA may still be active.

Validation:
- `python -m pytest --noconftest tests/test_shutdown_cuda_devices.py -q`: **18 passed**. CUDA calls are mocked; daemon threads and extracted worker/engine method bodies are real. Covers nonzero devices, TP deduplication, partial IPC failure, host registration, selection/sync/thread-start failures, timeout, retained buffers, idempotence, and main/SWA engine cleanup.
- Changed Python files pass `py_compile`; `git diff --check` passes.
- Ruff passes on the new helper and tests. Touched legacy files retain the same 9 diagnostics as the base, with no new diagnostics.
- GPU runtime validation has not been performed. The macOS host has no CUDA/ROCm runtime; `--noconftest` avoids the repository-wide fixtures that load the native runtime.


XingLiu1:
GPU validation completed on 2026-09-10 at PR head `67087dfcd02ed749427befb402924b5bd9bfb31a`.

Environment: NVIDIA H20, PyTorch `2.13.0+cu130`, CUDA 13.0, Python 3.12. Built `c_ext.so` from this exact source and imported the complete production worker/TransferEngine modules. Verified the tested Python file hashes against the PR checkout.

- Reproduced both original drain omissions: five trials per call site, with the caller on GPU 3. The fresh thread synchronized GPU 0 and returned while GPU 3's completion event was still pending. The control explicitly drained GPU 3 before freeing buffers.
- All 10 GPU shutdown regression cases passed: GPU 0, GPU 3, GPU 3+5 with duplicate IDs; real 32 MiB H2D/D2H copies; device timeout while another device drains; unregister only after DMA completion; retention on timeout; repeated shutdown; main/SWA device collection.
- All 10 spawned-worker lifecycle cases passed: five single-GPU and five TP2 workers on GPUs 3/5, real TensorSharedHandle IPC and native CE transfers. Each worker performed 20 round trips, for 200 total, with elementwise equality checks. Each then shut down with a real D2H copy outstanding; completion and host-region cleanup were verified.

These were actual GPU operations, not mocked CUDA or AST-extracted methods. The engine case exercises its production shutdown/finally path with minimal engine state; the IPC cases instantiate real GPUCPUTransferWorker/tpGPUCPUTransferWorker classes.

Both the test container and all its GPU processes were stopped afterward. This validates the two shutdown fixes on CUDA; it is not KLX/PyTorch 2.9 segfault reproduction, a full model-serving test, or a soak test. No further production-code change was needed.

flexkv/transfer/worker.py
```diff
@@ -347,6 +365,8 @@ def _register_host_tensor(self, tensor: torch.Tensor, label: str = "") -> None:
         )
         cudaHostRegister(tensor)
         self._host_registered.append((tensor, label or "host"))
+        # CPU-only workers can initialize CUDA through host registration too.
+        self._cuda_device_ids.add(torch.cuda.current_device())
```
linhu-nv, line 369: revise to 
```
 if torch.cuda.is_initialized():                                                                                                                                                                                                                                              
      self._cuda_device_ids.add(torch.cuda.current_device())  
```
because not all workers need gpu context. Otherwise we will have extra unnecessary HBM usgae,

linhu-nv (approved):
LGTM, please address my comments, then we can merge. thanks

PR #297, opened 2026-09-17
add transfer durations and fix SIGINT/SIGTERM exit

zhjc1124:
A lost GPU registration left an orphan looping forever: the wait had no deadline and the subprocess ignored SIGINT. Ctrl+C was broken on healthy runs too -- once ready the child restored SIG_IGN, and as a non-daemon process it stranded the parent in multiprocessing's atexit join().

- registration wait raises after 300s with registered/expected counts
- SIGINT/SIGTERM take the graceful-exit path for the whole process lifetime
- is_ready() raises instead of hanging when the subprocess died during startup
- shutdown() tolerates a half-initialized transfer engine
- poll interval 1ms -> 50ms

Add wait / xfer / e2e histograms for completed transfer ops, labelled by transfer_type (0.5ms - 30s buckets), gated on FLEXKV_ENABLE_METRICS.


linhu-nv (approved):
LGTM

